Issue #41, opened 2026-07-10
Task Proposal: Right-hand 3D joint trajectory from egocentric clips (agentic_vbench_understanding)


M4D-SC1ENTIST:
# Task Proposal — Right-Hand 3D Joint Trajectory From Egocentric Clips

**Family:** `agentic_vbench_understanding`
**Proposed task id:** `hands-bimanual-manipulation-3d-joint-trajectory`

## One-line

Given three short egocentric (head-mounted) clips of a person manipulating tabletop
objects with both hands, reconstruct the **right hand's 20 skeleton joints as metric 3D
points in the camera frame** at a set of query frames.

## The videos and question

- Three ~2-minute first-person RGB clips (1408×1408, ≥720p), each showing bimanual
  tabletop manipulation. The camera model (projection + intrinsics + device extrinsic)
  for each clip is provided to the agent as `cameras.json`.
- For 12 query frames per clip (36 total), output the right hand's 20 canonical joints
  as `[x, y, z]` in metres in that clip's RGB camera frame.
- Multi-clip task → exempt from the single-video 10-minute length floor (per the
  family README: "a comparison task may instead use a pair (or a few) short clips").

## Why it's hard (target: strong agent < 0.10, > 50 tool-call turns)

- The answer is **metric 3D structure**, not any on-screen readout. Recovering a joint's
  position in metres from a single monocular fisheye view requires reasoning about the
  supplied camera geometry and the hand's motion/parallax across many neighbouring
  frames — there is no depth in one frame.
- Two hands overlap and occlude; the agent must consistently isolate the **right** hand
  across the whole clip.
- Scored by soft-PCK at a **3 cm** tolerance per joint. Cm-level accuracy over 720
  (frame×joint) units means an agent has to genuinely triangulate, not eyeball.

## Ground truth & scoring (deterministic, pure code)

- GT: the capture rig's logged hand-tracking, forward-kinematicked to 20 joints and
  expressed per clip in the camera frame (tier: logged).
- Scorer: `reward = mean over (clip, frame, joint) of clip(1 - ||pred-gt|| / 0.03, 0, 1)`.
  Pure numpy, no VLM/LLM. Oracle = 1.0; empty/null = 0.0; random ≈ 0.

## Anti-shortcut

- single_frame, no_media, frame_dump_no_tools all score ≈ 0 in simulation (no way to get
  metric 3D from one frame, a caption, or eyeballed 2D). Audio unused (n/a for the
[...]

PR #45, opened 2026-07-11
Add understanding task: right-hand 3D joint trajectory from egocentric clips

M4D-SC1ENTIST:
## What this adds

A new task under `agentic_vbench_understanding`: `hands-bimanual-manipulation-3d-joint-trajectory`.

The agent gets three short egocentric clips of a person manipulating small objects with both hands, plus a pinhole `cameras.json` and a `queries.json`. For 12 query frames per clip (36 total) it must output the right hand's 20 skeleton joints as 3D points in that clip's camera frame, in metres. Scoring is a soft PCK: each joint earns `clip(1 - d/0.03, 0, 1)` on its L2 error in metres, averaged over all clip/frame/joint units. Pure numpy, no model judge.

Related proposal issue: #41.

## Why it clears the bar

The answer is metric 3D structure, not anything visible in a single frame or on screen, so an agent has to work the camera geometry and hand motion across many frames to recover joints to a few centimetres.

Calibration (see `calibration/scores.md` and the four transcripts under `calibration/rollouts/`):

|---|---|---|
| oracle | 1.0 | - |
| empty / null | 0.0 | - |
| Claude Code (Opus 4.8) | 0.0 | 67 |
| Codex (GPT-5.5) | 0.0 | 13 |
| Antigravity | 0.0 | 45 |
| Cursor (Composer) | 0.0 | 57 |

Every agent scored 0.0. The long-horizon property is shown by Claude, Antigravity, and Cursor running 45 to 67 tool-call turns; Codex stopped on its own after 13 turns at the same 0.0, which is the agent ending its run rather than the task lacking work.

Anti-shortcut ablations are all at or below 0.15: single frame 0.0, no media 0.0, tool-less frame dump 0.004. Audio is not used.

## Media hosting

The clips plus the camera and query JSON are hosted on Hugging Face and baked into the task image at build time with a pinned URL and SHA256, the same pattern the v1.0 families use. The clips are rectified to a generic pinhole view so the intrinsics carry no capture-rig identifiers.

## Test plan

- `python3 scripts/understanding/check_task.py --task-dir tasks/agentic_vbench_understanding/hands-bimanual-manipulation-3d-joint-trajectory` passes the structure check.
- Building the image and running setup, `solve.sh`, then `judge.py` in Docker gives oracle reward 1.0 (verified pulling the media from Hugging Face).


christine1729:
Thanks for the PR. The task idea is good and the scorer looks right. Fixes before merge:

- [ ] Move the answer key out of the agent's image (`/baked/ground_truth.json`) to the verifier side under `steps/solve/tests/`, like the reference task.
- [ ] Re-run calibration in an isolated env that matches the image (same libraries, harness, and model).
- [ ] Run the SPEC section 8 ablations for real, in place of the simulated numbers.
- [ ] Provide a download link that fetches the exact clips with the same spec and metadata, pinned with SHA256.

Smaller:
- Some partial-solve evidence would help show the 3 cm target is reachable, since all agents scored 0 and the oracle copies the key.
- `judge.py` should reject NaN/Inf in `joints_m`.
- SPEC nits: `input.url` is not a real URL, and resolution says 1024 in one place and 1408 in another.

Happy to re-review once these are fixed.

M4D-SC1ENTIST:
Thanks for the review. Pushed a revision that covers all of it.

- Answer key off the agent image: the Dockerfile now bakes only the clips plus cameras.json and queries.json, each pinned with SHA256. ground_truth.json lives on the verifier side, in steps/solve/tests/ for the grader and a copy in steps/solve/solution/ for the oracle, which Harbor mounts only during those steps (/tests and /solution). solve.sh reads /solution/ground_truth.json instead of a baked path. I checked inside the built image that no ground_truth exists anywhere the agent can reach, and the oracle still scores 1.0 end to end.
- Ablations are now real Claude Code runs on degraded inputs, in place of the simulated numbers: single_frame 0.0 (43 turns of trying to triangulate from one frame per clip), no_media 0.0, frame_dump_no_tools 0.0. Transcripts and summaries are under calibration/ablations/. These ran on the host where general CV libraries are available, so the shipped image (numpy plus ffmpeg only) can only score lower.
- Partial-solve evidence: added a partial-credit curve to scores.md. A method accurate to 5 mm scores 0.73, 15 mm scores 0.25, 20 mm scores 0.15. The reward rises smoothly, so the all-zero agent results reflect difficulty rather than an unreachable bar.
- judge.py now rejects non-finite joints_m (NaN or Inf scores 0 for that frame).
- SPEC nits: input now gives a real base_url with a per-file SHA256 list, and the 1024 vs 1408 mismatch is fixed (the clips are 1024 pinhole).

The download link point is handled by the pinned per-file URLs in the Dockerfile and the SHA256 list in SPEC section 9. Media is on Hugging Face, same host as the v1.0 families.

christine1729:
Hi @M4D-SC1ENTIST, thanks for the careful rev!

before merge:
- [ ] One fresh run with a current model (Fable 5 or GPT 5.6 Sol) on the shipped task
- [ ] Fix the stale task.toml comment that still says clips and GT are baked at build time

M4D-SC1ENTIST:
Both points are in.

- Fresh run with a current model: I ran Claude Code with Fable 5 on the shipped task, inside the built task image itself, so the environment matches the shipped Dockerfile exactly and only the baked materials were reachable. The container needs network for the model API, which also lets the agent pip install packages: it installed OpenCV and MediaPipe Hands, ran a real hand-landmark detector with multi-crop passes and a handedness disambiguation step, and submitted a complete, well-formed 36-frame solution. It still scored 0.0, with none of the 720 units inside 3 cm. That environment is strictly stronger than the shipped image (numpy plus ffmpeg, no network), so the in-image score cannot be higher. The transcript is `calibration/rollouts/claude-code-fable.jsonl` and runs to the CLI's closing result record (num_turns 66, of which 64 are tool calls).
- task.toml: the stale comment claiming clips and GT are baked at build time now says only the clips and agent-facing metadata are baked, and the ground truth ships verifier side.

Two related things I fixed while rechecking: the public HF dataset still held a ground_truth.json from before the key moved verifier side, which I deleted (the pinned build pulls only the materials files, so builds are unaffected), and the Antigravity row in the calibration table now names its model (Gemini 3.5 Flash), matching what the other two PRs were asked for.

techgenmini:
> Thanks — I rechecked the current head. The earlier GT-isolation, SHA-pinned materials, real-ablation, non-finite handling, and SPEC fixes are in. The August stale-comment fix, current-model trace, and
  > Antigravity label are also present.
  >
  > Two new findings need resolution before the original calibration-parity item can close:
  >
  > - [ ] Align the camera-coordinate convention with GT and add a projection regression. The prompt says +Y is down and uses v = fy*Y/Z + cy, but current GT uses the opposite Y sign. A prompt-compliant answer
  >   is therefore vertically flipped relative to the judge.
  >
  > - [ ] Make the metric targets observable, then do one final Fable run in the final shipped configuration. cameras.json supplies only intrinsics, while the GT is derived using a logged device trajectory and
  >   factory extrinsics; no metric reference, trajectory, or hand-scale calibration is agent-visible. The Fable trace also synthesizes/corrects four required non-fully-visible targets (clip_01 492/1737;
  >   clip_03 1148/2193) and reports 10–15% depth uncertainty against a 3 cm tolerance. Please provide sufficient agent-visible metric geometry and use fully visible query frames, or rescope the task/metric.
  >   The oracle-noise curve shows scorer smoothness, not recoverability from these inputs.
  >
  > After that correction, please run Fable once under the actual shipped environment—or deliberately update that environment and calibrate that version. The current run installed OpenCV and MediaPipe over the
  > network despite allow_internet=false and a numpy/ffmpeg image; no additional model campaign is requested.
  >
  > - [ ] Under the rollout-record policy, publish the complete Fable trajectory as an immutable archive with a whole-file SHA256. The committed trace elides 111 frame payloads. Mechanical secret/path
  >   redactions are fine; this is an archive correction, not a rerun.

tasks/agentic_vbench_understanding/hands-bimanual-manipulation-3d-joint-trajectory/steps/solve/instruction.md
```diff
@@ -0,0 +1,76 @@
...
+hand's** 20 skeleton joints, expressed **in that clip's RGB camera coordinate frame, in
+metres**.
+
+You are also given, in `/workspace/materials/`:
+
+- `cameras.json`: the pinhole camera intrinsics for each clip, keyed by clip id
+  (`clip_01`, `clip_02`, `clip_03`). Each entry gives `model` (`pinhole`), the image
+  `width`/`height`, the focal lengths `fx`/`fy`, and the principal point `cx`/`cy`
+  (standard pinhole model: a 3D camera-frame point `[X,Y,Z]` projects to pixel
+  `u = fx*X/Z + cx`, `v = fy*Y/Z + cy`).
```
techgenmini, line 20: > The stated +Y-down / v = fy*Y/Z + cy convention does not match current GT. For clip_01:195, the GT wrist projects to (988, 189) with the prompt formula, while the visible wrist is lower-right; the opposite-Y formula gives (988, 834).
  > Please align the convention and add a GT projection-overlay regression.

M4D-SC1ENTIST:
All four are in, and the convention finding led somewhere bigger than a sign flip.

Chasing your Y-axis observation, I reprojected the baked joints onto the pinned frames and found the ground truth was expressed in the native camera frame while the shipped clips are rectified with an upright 90 degree rotation. So the prompt's convention was right for the video, and the GT was wrong for both: a correct answer read off the video failed the 3 cm test by construction, which retroactively explains the all-zero table. The GT is regenerated in the frame the video and cameras.json actually describe, verified by projecting joints onto the pinned frames (they now land on the hand). `tests/test_gt_projection.py` locks this in: every query must project in-image under the exact formula the instruction states, and GT bone lengths must match the shipped hand model. Re-scoring the earlier Fable submission against its own frames with only the frame fix applied moves it from 0.0 to 0.036, so the task now rewards correct work.

On observability: materials/ now ships `hand_model.json`, the wearer's rigid bone lengths (constant across all GT frames; the one non-rigid wrist-thumb link is omitted). Bone pixel span plus the pinhole intrinsics fixes metric depth, so the target is recoverable from agent-visible inputs alone. The four queries you flagged are gone: queries are re-picked with a projection test requiring all 20 joints inside the image with a margin, since the rig's visibility mask refers to the much wider pre-rectification FOV.

The final Fable run is on the corrected task under the shipped configuration: inside the built task image, network restricted by a DNS allowlist so only the model endpoint resolves (its two pip attempts failed, matching allow_internet=false; the model channel is harness-side in Harbor), 60 minute budget, shipped tools only. It scored 0.041 in num_turns 55 and finished on its own in 32 minutes, working from hand-annotated 2D joints plus the bone-length scale. Ablations were re-run the same way: single_frame 0.009, no_media 0.0, frame_dump 0.0. So the scale reference works, the bar stays out of reach of shortcuts, and 0.041 sits well under 0.10.

Archive: complete transcripts of every retained trajectory, frames included, are at a revision-pinned dataset URL with per-file SHA256 in `calibration/scores.md`, same policy as the other two PRs. The scrub-script question is being settled on the reconstruction PR; the same exception request covers this archive.

Issue #42, opened 2026-07-10
Task Proposal: Object 6DoF pose trajectory from egocentric clips (agentic_vbench_understanding)


M4D-SC1ENTIST:
# Task Proposal — Object 6DoF Pose Trajectory From Egocentric Clips

**Family:** `agentic_vbench_understanding`
**Proposed task id:** `objects-tabletop-manipulation-6dof-pose-trajectory`

## One-line

Given three short egocentric clips of a person picking up and turning over tabletop
objects, reconstruct a **designated object's full 6DoF pose (position + orientation) in
the camera frame** at a set of query frames.

## The videos and question

- Three ~2-minute first-person RGB clips (1408×1408, ≥720p). Each clip designates one
  moving target object (`objects.json`); the camera model is provided (`cameras.json`).
- For 12 query frames per clip (36 total), output the object's pose as translation
  `t_xyz_m` (metres, camera frame) and orientation `q_wxyz` (unit quaternion).
- Multi-clip task → exempt from the single-video length floor.

## Why it's hard (target: strong agent < 0.10, > 50 tool-call turns)

- A metric 6DoF pose from a monocular fisheye clip demands both **metric depth** and
  **3D orientation** — each resolves only by integrating the object's motion/parallax
  across many frames under the supplied camera model.
- The object is grasped, lifted, and rotated: its camera-frame pose changes a lot
  (net displacement ~0.4–1.9 m across a clip), so a single guessed pose cannot cover
  the trajectory.
- Scored by soft-ADD (BOP's ADD metric) at 10% of the object diameter. Getting the
  translation right but the rotation wrong scores 0 — the full pose must be recovered.

## Ground truth & scoring (deterministic, pure code)

- GT: the rig's logged per-frame object pose, transformed into the camera frame
  (tier: logged). All three target objects are asymmetric (birdhouse_toy, vase,
  potato_masher), so orientation is well-defined.
- Scorer: per query, ADD = mean over the object's mesh points of `||T_pred·p − T_gt·p||`;
  `frame_score = clip(1 − ADD/(0.1·diam), 0, 1)`; reward = mean over 36 queries. Pure
  numpy. Oracle = 1.0; empty/null = 0.0.

## Anti-shortcut

- single_frame ≈ 0.08, no_media = 0, frame_dump_no_tools = 0 in simulation. Audio
  unused (n/a).

Built to the Harbor layout and calibrated locally (oracle 1.0, null 0, ablations ≤ 0.083).
[...]

PR #46, opened 2026-07-11
Add understanding task: object 6DoF pose trajectory from egocentric clips

M4D-SC1ENTIST:
## What this adds

A new task under `agentic_vbench_understanding`: `objects-tabletop-manipulation-6dof-pose-trajectory`.

The agent gets three short egocentric clips, a designated target object per clip (in `objects.json`), a pinhole `cameras.json`, and a `queries.json`. For 12 query frames per clip (36 total) it must output the object's 6DoF pose in the camera frame: translation in metres plus a unit quaternion. Scoring is the ADD metric (Hinterstoisser / BOP): the object's baked mesh points are transformed by the predicted and ground-truth poses, and the mean point distance is mapped to `clip(1 - e/(0.1*diam), 0, 1)`. Pure numpy, no model judge.

Related proposal issue: #42.

## Why it clears the bar

A metric 6DoF pose needs both depth and 3D orientation, which only resolve by tracking the object across many frames under the given intrinsics. Getting the translation right but the rotation wrong still scores 0, so the whole pose has to be recovered.

Calibration (see `calibration/scores.md` and the transcripts under `calibration/rollouts/`):

|---|---|---|
| oracle | 1.0 | - |
| empty / null | 0.0 | - |
| Claude Code (Opus 4.8) | 0.0 | 80 |
| Codex (GPT-5.5) | 0.0 | 12 |
| Antigravity | 0.0 | 100 (isolated) |
| Cursor (Composer) | 0.0 | 62 |

Every agent scored 0.0. Claude, Antigravity, and Cursor ran 62 to 100 turns; Codex stopped on its own after 12 turns at the same 0.0.

One integrity note worth flagging: Antigravity is an aggressive retriever. On an earlier host run it crawled the filesystem and web to recognise the source data and copy answers instead of solving the vision task. I therefore ran it in a filesystem-isolated Docker container (network on for the model, but no repo, no ground truth, and no dataset files reachable), and the clips are rectified to a generic pinhole with all capture-rig identifiers stripped from the intrinsics. With no way to retrieve the answer it produced no valid pose, the honest 0.0. The other agents were confirmed clean from their transcripts.

Anti-shortcut ablations are at or below 0.15: single frame 0.083, no media 0.0, tool-less frame dump 0.0. Audio is not used.

## Media hosting

Clips plus camera, query, and object JSON are hosted on Hugging Face and baked into the task image at build with a pinned URL and SHA256, the same pattern the v1.0 families use.

## Test plan

- `python3 scripts/understanding/check_task.py --task-dir tasks/agentic_vbench_understanding/objects-tabletop-manipulation-6dof-pose-trajectory` passes the structure check.
- Building the image and running setup, `solve.sh`, then `judge.py` in Docker gives oracle reward 1.0 (verified pulling the media from Hugging Face).


christine1729:
Thanks for the PR. The scorer itself is fine, but there is one design problem plus the same setup issues as #45.

The big one:
- [ ] The task looks impossible, not just hard: the pose is defined in the object's mesh frame but the agent never gets the mesh, so even perfect vision cannot land inside the ADD threshold, and every agent scoring exactly 0.0 points the same way. Please ship the mesh points (or define the frame another way) in `materials/`.

Same as #45, see the notes there:
- [ ] Answer key out of the agent's image (here it also sits in the same public HF dataset as the materials, so please separate those too).
- [ ] Calibration in an isolated env that matches the shipped task.
- [ ] Real ablations for SPEC section 8.
- [ ] A pinned download link for the exact clips with the same spec and metadata.
- [ ] Same SPEC nits (url, 1024 vs 1408).

Nit: one sentence in SPEC justifying three 2-minute clips under the comparison-pair exemption would help.

Once the mesh ships and a real agent can score above 0, this becomes a strong task.

M4D-SC1ENTIST:
Thanks, the impossibility call was right. Pushed a revision.

- Root cause was that the pose was defined in the object's canonical frame but the agent never got that frame, so even a perfect estimate could not land inside ADD. materials/ now includes object_points.json: the canonical point set the pose maps from, which is also the exact set the ADD metric scores on. With both the points and the camera model, a correct pose is well defined. The instruction now explains this and I removed the line that wrongly said the model was not needed.
- To show the bar is reachable, scores.md has a partial-credit curve from perturbing the oracle poses: 5 mm / 2 deg scores 0.71, 10 mm / 5 deg scores 0.43, 20 mm / 8 deg scores 0.13. So a good model-based pose estimate is rewarded well before it is exact; the difficulty is getting metric depth and orientation from a single moving view.
- Answer key is off the agent image, same setup as #45: the Dockerfile bakes only the clips plus cameras.json, queries.json, objects.json, object_points.json, each SHA256-pinned. The poses ship verifier side, in tests/ for the grader and a solution/ copy for the oracle. solve.sh reads /solution/ground_truth.json.
- Ablations are real Claude Code runs: single_frame 0.0 in 49 turns (it had the object points and one frame per clip and still could not fix depth), no_media 0.0. Transcripts in calibration/ablations/.
- SPEC input has a real base_url with a per-file SHA256 list, and a sentence on why three short clips use the comparison exemption.

Same HF host as the v1.0 families for the media.

christine1729:
Hi @M4D-SC1ENTIST, thanks for the rev!

before merge:
- [ ] One complete fresh run with a current model on the fixed task (the current Claude rollout is truncated mid-run with no result record)
- [ ] Name the Antigravity model in the table; the row does not say which Gemini it was, so the run cannot be tied to a listed model.
- [ ] Sync the 49-vs-50 turn count between the summary and the transcript.

M4D-SC1ENTIST:
All three are in.

- Fresh complete run: Claude Code with Fable 5 on the fixed task, executed inside the built task image. The transcript (`calibration/rollouts/claude-code-fable.jsonl`) runs to the CLI's closing result record: num_turns 226, of which 225 are tool calls, over 98 minutes. The agent used object_points.json to build metric landmark models for all three objects, hand-annotated landmarks per query frame, fit poses with PnP, verified them by reprojecting the full point cloud onto crops, and covered all 36 queries with plausible tabletop depths. Every frame's ADD still exceeded the 0.1-diameter tolerance, so it scored 0.0. Metric depth from a single moving view is what keeps it out; the partial-credit curve in scores.md shows the bar itself is reachable. The old truncated Opus transcript is removed and the Fable row replaces it.
- Antigravity model: it was Gemini 3.5 Flash, now named in the table. scores.md also gained explicit tables tying each rollout file to a row. The codex, cursor, and antigravity rollouts predate object_points and are labeled as evidence for the unsolvable-by-design diagnosis of the earlier revision, not as rows on the current design.
- Turn count: the 49 was my count of tool_use blocks; the transcript's own result record says num_turns 50. The docs now quote the result record and say that is the metric.

While rechecking hosting I also found the public HF dataset still held a ground_truth.json from before the key moved verifier side. Deleted; the dataset now carries only the agent-facing materials and the pinned build is unaffected.

techgenmini:
> Thanks — the August follow-up checklist is complete, and I verified the earlier object-frame, GT-isolation, ablation, and media-pinning fixes.
  >
  > One item from the original review remains:
  >
  > - [ ] The new Fable trace is complete, but it is not calibrated on the shipped configuration: it ran for 96.6 minutes and installed Pillow/OpenCV via pip, while the task declares a 1-hour,
  >   allow_internet=false image that does not include those packages. Please run once under the shipped configuration, or deliberately update that configuration and calibrate the revised task. No additional
  >   model campaign is requested.
  >
  > New finding from checking the retained trace against the pinned clips:
  >
  > - [ ] Five required queries do not show the designated target: clip_01 667/2317/3009, clip_02 2943, and clip_03 473. The trace describes a rotation-only transfer for them and estimates 10–30 cm uncertainty,
  >   versus a 2.4–3.5 cm ADD tolerance. Please replace them with target-visible frames and regenerate GT, or provide sufficient camera/object-motion signal to support that transfer. The parity run above should
  >   use the corrected task.
  >
  > Separately, under the rollout-record policy:
  >
  > - [ ] Publish the complete Fable trajectory as an immutable archive with URL and whole-file SHA256. The committed trace has 426 elided frame placeholders. Allowed mechanical path/secret redactions are fine;
  >   this is an archive correction, not a rerun.

tasks/agentic_vbench_understanding/objects-tabletop-manipulation-6dof-pose-trajectory/steps/solve/tests/judge.py
```diff
@@ -0,0 +1,135 @@
...
+    P = (R_p @ pts.T).T + t_p
+    G = (R_g @ pts.T).T + t_g
+    if symmetric:
+        from scipy.spatial import cKDTree
+        d, _ = cKDTree(G).query(P)
+        return float(d.mean())
+    return float(np.linalg.norm(P - G, axis=1).mean())
+
+
+def _index_pred(pred):
```
techgenmini, line 55: > int(e["frame"]) is outside the error handling, so "frame": "bad" or null aborts scoring instead of receiving the documented zero credit for malformed entries. Please skip invalid frame IDs and add a
  > regression test; this does not require recalibration.

M4D-SC1ENTIST:
All four are in (the three from your comment plus the judge.py inline note), and the visibility finding led to a bigger fix.

The five frames first. Chasing why they slipped through, I reprojected the baked poses onto the pinned frames and found the ground truth was expressed in the native camera frame while the shipped clips are rectified with an upright 90 degree rotation. So there were two problems stacked: the flagged queries never showed the target in the rectified FOV (the rig's visibility mask refers to the much wider pre-rectification view), and even on visible frames a correct pose read off the video failed ADD by construction. That second part retroactively explains why every careful run, including the 226-turn Fable one, scored exactly 0.0. Fixes: poses are regenerated in the frame the video and cameras.json describe, queries are re-picked with a projection test requiring at least 98 percent of the object points inside the image with a margin, and ADD still scores on exactly the object_points.json set. `tests/test_gt_projection.py` locks the convention as a regression. Re-scoring the earlier Fable submission against its own frames with only the frame fix applied moves it from 0.0 to 0.045, with individual frames earning up to 0.52, so the task now rewards correct pose work.

The parity run is on the corrected task under the shipped configuration: inside the built task image, network restricted by a DNS allowlist so only the model endpoint resolves (pip fails, matching allow_internet=false; the model channel is harness-side in Harbor), 60 minute budget, ffmpeg plus numpy only. Complete transcript with the closing result record, num_turns 122: score 0.002. Without a PnP solver it annotated landmarks by eye and hand-rolled the geometry; one frame reached 0.069. Ablations were re-run the same way on the fixed task: single_frame 0.0 in 77 turns, no_media 0.0.

judge.py: frame ids now go through a guarded int conversion, so "frame": "bad" or null skips that entry with zero credit instead of aborting, with regression tests in `tests/test_judge.py` covering malformed ids, malformed pose payloads, and non-dict shapes.

Archive: complete transcripts of every retained trajectory, frames included, are at a revision-pinned dataset URL with per-file SHA256 in `calibration/scores.md`. The 426-placeholder trace you flagged is there in full as `claude-code-fable-opennet.full.jsonl`. The scrub-script policy question is being settled on the reconstruction PR; the same exception request covers this archive.

Issue #43, opened 2026-07-10
Task Proposal: 3D object reconstruction from egocentric manipulation clips (agentic_vbench_understanding)


M4D-SC1ENTIST:
# Task Proposal — 3D Object Reconstruction From Egocentric Manipulation Clips

**Family:** `agentic_vbench_understanding`
**Proposed task id:** `object-reconstruction-from-egocentric-manipulation`

## One-line

Given three short egocentric clips, each showing one tabletop object being picked up and
rotated in hand, reconstruct the **3D surface mesh** of each clip's designated object.

## The videos and question

- Three ~2-minute first-person RGB clips (1408×1408, ≥720p). Each designates a target
  object (`objects.json`); the camera model is provided (`cameras.json`).
- Output one triangle mesh per clip (`clip_01.obj` / `.ply` / `.glb` / `.stl`, etc.).
- Scale and pose are free — the mesh is scored on **shape** after a best-fit similarity
  alignment. Multi-clip task → exempt from the length floor.

## Why it's hard (target: strong agent < 0.10, > 50 tool-call turns)

- The full 3D surface only emerges by fusing many viewpoints across the clip as the
  object is turned over; no single frame shows the whole surface (self-occlusion), and
  no stock model may be substituted.
- Three separate objects across three clips — each must be reconstructed from its own
  clip.
- Scored by F-score at a strict fraction of the object diameter after alignment, so a
  coarse blob does not pass; the surface geometry (including concavities) must match.

## Ground truth & scoring (deterministic, pure code)

- GT: the scanned 3D reference mesh of each interacted object, sampled to a dense point
  cloud (tier: logged).
- Scorer: per clip, sample the submitted + reference meshes, best-fit similarity-align
  (PCA-frame + octahedral restarts + similarity-ICP), then a shape-agreement score;
  reward = mean over the three clips. Pure numpy/scipy/trimesh, no VLM/LLM. Oracle = 1.0
  (any pose/scale of the correct shape); empty = 0.0; wrong object ≈ 0.

## Anti-shortcut

- no_media = 0; wrong-object (e.g. a keyboard mesh) ≈ 0; a convex hull or extruded
  silhouette slab (the best a single frame or tool-less agent can do) is driven below
  the bar by the concavity-sensitive metric and by choosing genuinely non-convex target
  objects. Audio unused (n/a).

Built to the Harbor layout and being calibrated locally (oracle 1.0, null 0). Requesting
a sanity-check before I open the PR.

PR #47, opened 2026-07-11
Add understanding task: 3D object reconstruction from egocentric manipulation clips

M4D-SC1ENTIST:
## What this adds

A new task under `agentic_vbench_understanding`: `object-reconstruction-from-egocentric-manipulation`.

The agent gets three short egocentric clips, each showing one target object being picked up and turned over, plus a pinhole `cameras.json` and an `objects.json`. It must reconstruct a 3D surface mesh of each clip's object and write one mesh per clip (`clip_01.obj` and so on; obj, ply, glb, or stl accepted). Scale and pose are free: each mesh is compared to a reference after a best-fit similarity alignment, then scored by a surface F-score squared times a flood-filled volumetric occupancy IoU, normalised per object so the true shape scores 1.0. Pure numpy, scipy, and trimesh, no model judge.

Related proposal issue: #43.

## Why it clears the bar

The full surface only emerges by fusing many viewpoints as the object is rotated; no single frame shows the whole shape. The occupancy term is what makes the task shortcut-resistant. A convex hull fills real concavities and a flat silhouette slab collapses the volume, so both score well below the bar, while a wrong object scores near zero.

Calibration (see `calibration/scores.md` and the transcripts under `calibration/rollouts/`):

|---|---|---|
| oracle | 1.0 | - |
| empty / null | 0.0 | - |
| wrong object (keyboard mesh) | <= 0.002 | - |
| Claude Code (Opus 4.8) | 0.043 | 46 |
| Codex (GPT-5.5) | 0.014 | 11 |
| Antigravity | 0.033 | 64 |
| Cursor (Composer) | 0.019 | 60 |

Every agent scored below 0.10. Claude, Antigravity, and Cursor ran 46 to 64 turns reconstructing three meshes each; none matched the true surface closely enough. Codex stopped on its own after 11 turns.

Anti-shortcut ablations at or below 0.15: an empty output is 0.0, a wrong object is 0.003 to 0.009, a silhouette slab is 0.005 to 0.012, and a convex hull (the best a tool-less guess can do) is 0.017 for the spatula and 0.127 for the coffee pot. Audio is not used.

## Media hosting

Clips plus camera and object JSON, and the reference meshes used by the oracle and verifier, are hosted on Hugging Face and baked into the task image at build with a pinned URL and SHA256, the same pattern the v1.0 families use.

## Test plan

- `python3 scripts/understanding/check_task.py --task-dir tasks/agentic_vbench_understanding/object-reconstruction-from-egocentric-manipulation` passes the structure check.
- Building the image and running setup, `solve.sh`, then `judge.py` in Docker gives oracle reward around 0.999 to 1.0 (verified pulling the media from Hugging Face).


christine1729:
Thanks for the PR. The scorer is well designed and holds up against hull and slab tricks.

Same as #45, see the notes there:
- [ ] Answer key out of the agent's image (here the `ref_clip_0*.glb` meshes, which the oracle itself copies for a 1.0), and please also show the task is solvable above the current best of 0.043.
- [ ] Real ablations, with turns reported for the strongest agent (Claude: 0.043 in 46 turns, under the 50-turn gate).
- [ ] A pinned download link for the exact clips and meshes with the same spec and metadata.

New in this PR:
- [ ] Seed the sampler: `_SAMPLE_SEED = 0` is defined but never used, so `trimesh.sample_surface` gives different scores each run near the 0.15 bar.
- [ ] Pin `cameras.json`, `objects.json`, and the ref meshes with SHA256 like the clips, so the task cannot change after merge.

Discussion:
- Use three different objects, since two clips share `coffee_pot` and one rebuild covers 2/3 of the reward.

M4D-SC1ENTIST:
Pushed the fixes. Rundown by point:

**Answer key out of the image.** The image now bakes only the three clips, cameras.json, and objects.json. The reference meshes live verifier side: a copy under `steps/solve/solution/` for the oracle, and the grader scores against the point set in `steps/solve/tests/ground_truth.json`. I built the real image and checked the agent phase directly: `find / -name 'ref_clip*' -o -name 'ground_truth*'` returns nothing, so no reference mesh or ground truth is anywhere in the image the agent runs in. Oracle end to end is still 1.0.

**Seed the sampler.** `_SAMPLE_SEED` is now passed to `sample_surface` (with a fallback to seeding numpy's global RNG for older trimesh). I graded the oracle output three times on the built image and got identical scores each run.

**Pin cameras.json and objects.json.** Both are now SHA256 checked at build like the clips. While doing this I noticed cameras.json was still the raw wide-angle intrinsics with capture metadata, which contradicted the instruction that promises a pinhole model. It now ships the stripped pinhole model (1024x1024, fx/fy 500), matching the other two families.

**Three different objects.** Now coffee_pot, potato_masher, spatula_red. I first tried a bowl for the second clip, but a bowl is a surface of revolution and barely rotates in its clip, so it was recoverable from almost a single frame and a strong agent scored 0.32 on it, over the bar. The potato masher has a thin handle and a perforated head, needs many viewpoints, and its convex hull scores 0.003.

**Solvable above the best agent.** Partial-solve check on the true meshes: decimating to 5% of the faces still scores 1.0 (a coarse but correct surface passes), and perturbing every vertex by 2% of the diameter scores 0.116 mean, above the best real agent. So the bar is reachable; the gap is reconstruction accuracy.

**Real ablations and turns.** All numbers are from real runs on the new set. Convex hull 0.046 / 0.003 / 0.006, silhouette slab about 0.006 each, both well under the bar. All four agents stay under 0.10 (Codex 0.034, Claude 0.018, Cursor 0.006, Antigravity 0.001), and three of four run well past 50 turns (Antigravity 153, Cursor 116, Claude 52), so the low scores are the difficulty, not early stopping. Antigravity built a full COLMAP structure-from-motion plus Poisson meshing pipeline and still landed at 0.001. I ran it in a filesystem-isolated container that mounts only the materials, so it could not look at other runs. Transcripts refreshed in `calibration/rollouts/` (base64 frame payloads elided to keep sizes down).

**Pinned download.** Clips and metadata are pulled from the Hugging Face dataset in the Dockerfile with a SHA256 for every file. The reference meshes are not hosted there; they ship in the repo verifier side.

christine1729:
Hi @M4D-SC1ENTIST, thanks for the rev!

before merge:
- [ ] One fresh run with a current model (all rows are GPT-5.5 era), and name the Antigravity model in the table.

M4D-SC1ENTIST:
Both are in.

- Fresh current-model run: Claude Code with Fable 5 on the shipped task, executed inside the built task image, so only the baked materials were reachable and the toolset matched the shipped Dockerfile (ffmpeg, numpy, scipy, trimesh; it installed nothing extra). It scored 0.041 in 101 tool calls (the closing result record says num_turns 102), which makes it the strongest attempt so far and still well under 0.10. Its approach: locate the manipulation segment in each clip, segment the object by color, measure per-view width profiles with a foreshortening correction across near-orthogonal views, then build each mesh procedurally from those measurements. Transcript: `calibration/rollouts/claude-code-fable.jsonl`. Table and SPEC updated; the solvability margin holds, since 2% vertex noise on the true meshes scores 0.116, above 0.041.
- Antigravity model named: Gemini 3.5 Flash.

Same hosting fix as the other two PRs: the HF dataset still carried the old ref_clip_*.glb meshes, which I deleted. The dataset now holds only the agent-facing materials; the Dockerfile never pulled the meshes, so the pinned build is unaffected.

techgenmini:
Hi @M4D-SC1ENTIST, thanks — the Aug 24 checklist is now closed. I verified that the Fable 5 run is a fresh 101-tool-call run within the shipped one-hour budget, using only the shipped image tools; no further calibration run is requested.

  One merge blocker remains:

  - [ ] `rollouts/README.md` says sampled-frame payloads were elided. Please publish complete versions of each retained calibration trajectory, including those frames, in an immutable archive (release/LFS or a
  revision-pinned dataset), and add the URL and whole-file SHA256 for each in `calibration/scores.md`. If a permitted redaction is needed, document the mechanical scrub process. This is an archive correction,
  not a rerun.

  Small documentation sync in the same update: `SPEC.md` says the surface tolerance is 4× median NN spacing, while the shipped judge uses 2.5× (and normalises F by the oracle ceiling). Please make the card
  match the code; no recalibration is needed.

M4D-SC1ENTIST:
Both are in.

- Archive: complete versions of all five retained trajectories, sampled frames included, are published to a revision-pinned HF dataset, so the URLs are immutable. `calibration/scores.md` now lists the base URL and the whole-file SHA256 of each file. The redaction is mechanical and documented in the archive repo's README: absolute host paths from the runner machine become `/workspace`, and capture-hardware identifiers become neutral phrases so the tasks cannot be reverse-searched from the transcripts. Nothing is removed or elided in the archive; the in-repo copies are the same files after base64 frame elision.
- SPEC: the scorer card now matches judge.py. Tolerance is 2.5 median NN spacings (TAU_SPACING_MULT), and the formula is written as (surface-F / oracle_f)^2 * (voxel-IoU / oracle_iou), both ratios clipped to 1, which is what the code computes. The 4x wording was left over from an earlier tuning pass.

techgenmini:
> Thanks — I rechecked the current head. The prior archive and SPEC asks are now satisfied: the Fable archive is revision-pinned, its SHA256 matches scores.md, and the complete file contains no elided frame payloads. The scorer, GT
  > isolation, and shipped-config Fable run also check out. No rerun is requested.
  >
  > One small POL-1 closure remains:
  >
  > - [ ] Please add or link the deterministic committed scrub script. The archive README says it replaces device/library/sequence/hosting-account identifiers with neutral phrases; those are outside the standard automatic redaction list.
  >   Either limit the script to approved redactions, or request a maintainer-approved task-integrity exception here for these specific source-identifying substitutions. This is an archive-policy decision, not a calibration issue.

M4D-SC1ENTIST:
On the scrub script: I would like to request a maintainer-approved task-integrity exception for the source-identifying substitutions, rather than committing the script itself.

The reason is that the substitution list is the leak. The replaced strings are the capture-device and tracking-library names, sequence identifiers, and the hosting account of the source collection; committing the script (or an approved-redactions-only variant that leaves them in) hands anyone, including a future agent under evaluation, a direct index from the transcripts to the collection's public annotations, which include this task's answer keys. This is not hypothetical: in an earlier calibration run on the sibling 6DoF task, an agent identified the source collection from metadata fingerprints alone and went looking for its published annotations before the isolation stopped it. Keeping the identifiers out of the public record is part of the same defense as the de-fingerprinted camera model.

What I can offer for auditability: the substitutions are a fixed list of deterministic regex replacements with neutral phrases, nothing is deleted, no line boundaries move, and payloads are untouched. I am happy to hand the exact script to a maintainer through any private channel you prefer so the archive can be re-derived and diffed against the raw logs, and to state in the archive README that the script is available to maintainers on request. If you would rather have a different mechanism (hash-committed script, sealed copy in a maintainer-only location), name it and I will set it up.

techgenmini:
Hi @M4D-SC1ENTIST, thanks for spelling out the threat model. I agree that publishing the substitution list would itself create the lookup path this task is trying to prevent, so I do not want the scrub script
  committed publicly.

  Let’s handle this as a scoped POL-1 task-integrity exception. Please send the exact deterministic scrub script, plus the raw and archived-file SHA256s, privately to a maintainer(@techgenmini  or @christine1729 ). We will verify that it makes
  only the stated source-identifying substitutions and preserves tool calls, outputs, turns, line structure, and frame payloads.

  After that private check, we’ll record the approved exception and its scope here. No rerun or additional public archive is requested.

M4D-SC1ENTIST:
Sent. Since GitHub has no direct messages, I put the material in a private repo and invited both of you as read collaborators: https://github.com/M4D-SC1ENTIST/avb-scrub-audit

It contains the exact substitution script (`scrub_rollout.py`), the entry point that produced every archive file (`make_full.py`, which reads a raw log, applies the substitutions, writes, and prints the sha256), and `SHASUMS.md` with the raw and archived sha256 for every file in the archive, covering all three task PRs so one check settles the same exception for 45 and 46. Verification is one command per file: `python3 make_full.py <raw_log> <out>` must reproduce the archived sha256 byte for byte. I ran that round trip myself on the package as shipped and it reproduces the published hashes.

One caveat, recorded in the manifest: the raw log of the objects antigravity run was overwritten when its container workspace got reused for a later run, so for that one file the archived copy is byte-identical to the in-repo committed one (that CLI logs plain narration with no image payloads, so nothing was elided there either). Raw logs for everything else are preserved, and I can add any of them to the same private repo if you want to re-derive rather than audit the script.

techgenmini:
> Hi @M4D-SC1ENTIST — thanks. I reviewed the existing private avb-scrub-audit repository. The scrubber is narrowly limited to source-identifying/path substitutions, and I verified that all five revision-
  > pinned #47 recon/ archive files match the SHA256s in scores.md.
  >
  > One POL-1 closure remains:
  >
  > - [ ] Please add the original raw inputs for the transformed Claude, Codex, and Cursor recon/ traces to that same private repository (or provide equivalent maintainer-only links). They do not need to be
  >   public. The manifest currently lets us verify claimed hashes, but not run the documented raw → make_full.py → archived-file round trip. Once that reproduces the listed hashes, we can record the scoped #47
  >   exception.
  >
  > Small packaging cleanup in the same update:
  >
  > - [ ] Replace the materials resolve/main URL with a Hugging Face revision hash, digest-pin python:3.12-slim, and remove the unused HF_TOKEN passthrough. These are reproducibility/security changes only; no
  >   rerun is requested.
  >
  > After that, I’m happy to approve.

tasks/agentic_vbench_understanding/object-reconstruction-from-egocentric-manipulation/task.toml
```diff
@@ -0,0 +1,33 @@
...
+
+[environment]
+build_timeout_sec = 2400.0
+cpus = 4
+memory_mb = 8192
+storage_mb = 8192
+allow_internet = false
+
+[environment.env]
+HF_TOKEN = "${HF_TOKEN:-}"
```
techgenmini: > These assets are public and baked during image build, so this host-token passthrough is unnecessary. Please remove HF_TOKEN from the task environment.

M4D-SC1ENTIST:
Both are done.

Raw inputs: the private repo now has `raw/recon/` with the original logs for the four transformed traces (both Claude runs, Codex, Cursor), so the documented round trip runs there directly: `python3 make_full.py raw/recon/<file> <out>` reproduces each archived SHA256, which I re-checked for all four before pushing. The recon antigravity raw is not included because the scrub does not touch it; its raw and archived hashes in SHASUMS.md are identical, which is the round-trip result for that file. The hands and objects raws are preserved too and I can add them the same way whenever you want to close the equivalent exception on 45 and 46.

Packaging: MATERIALS_BASE now resolves against the dataset revision hash instead of main, the base image is digest-pinned (python:3.12-slim by digest), and the HF_TOKEN passthrough is gone from task.toml since the assets are public and fetched at build time. Rebuilt from the pinned Dockerfile: every per-file SHA256 check passes and the oracle scores 1.0.

techgenmini:
> Hi @M4D-SC1ENTIST — rechecked. I ran the maintainer-only raw → make_full.py round trip for both Claude recon streams, Codex, and Cursor, then byte-compared each result with its revision-pinned archive file.
  > All four reproduce the SHA256s recorded in scores.md. The Antigravity archive is also byte-identical to its committed plain-text transcript.
  >
  > The Hugging Face revision pin, digest-pinned base image, matching SPEC reference, and removal of the unused HF_TOKEN are all in place. The prior scorer, GT-isolation, calibration, and archive checks remain
  > satisfied.
  >
  > I’m approving and recording a narrow #47/recon POL-1 task-integrity exception for these verified private, deterministic source-identifying substitutions. It applies only to the recon archives verified here
  > and does not automatically extend to other task PRs.

Issue #50, opened 2026-07-12
[Task Proposal] Soccer restart-outcome chains over a full match


Y-Yiqing:
# Task Proposal — Soccer Restart Outcome Chains

**Family:** `agentic_vbench_understanding`
**Proposed task id:** `soccer-restart-outcome-chains`

## One-line

Given one full 90-minute soccer broadcast, for **every visible ball restart** recover
`(t, restart_type, restart_team, outcome)` in chronological order. It is not event
detection: each entry requires attributing the restart to a team and following the
ensuing 15-30s of play to its outcome, over 150+ events across both halves.

## The video and question

- One low-profile **SoccerNet-v2** full-match broadcast (~90 min, both halves), from a
  minor league so no per-event public timeline exists. Single long video, no length-floor
  concern.
- The clip is re-encoded and metadata-stripped; the match identity and its
  `Labels-v2.json` are kept off the agent's image/network path (answer key unreachable at
  runtime).
- Output, using the closed vocabulary below, every **visible** restart instance in time
  order:
  - `restart_type` in {Throw-in, Corner, Direct free-kick, Indirect free-kick}
  - `restart_team` in {home, away}
  - `outcome` in {goal_within_30s, shot_within_15s, none}
- Answer format: `{"sequence": [{"t": 61.4, "restart_type": 1, "team": "home", "outcome": 0}, ...]}`

## Why it's hard (target: strong agent < 0.10, > 50 tool-call turns)

- A dead-ball frame gives none of the answer. Attributing the restarting team and
  resolving the 15-30s outcome forces cross-moment reasoning per event, compounded over
  150+ four-field tuples — not a repeatable visual template (this is the fix for the
  "restart detection is solved CV" concern: detection alone scores near zero here).
- The events span both halves; each needs the agent to localize the restart AND watch
  the following play, so the answer is distributed across the whole broadcast.

## Ground truth & scoring (deterministic, pure code, no LLM judge)

- **GT provenance is fully mechanical** — every field is a deterministic transform of
  SoccerNet-v2's published, multi-annotator `Labels-v2.json` (per-event fields:
  `position` ms, `label`, `team`, `visibility`). No hand labeling, no single-annotator
  trap:
  - `t` / `restart_type` / `restart_team` = the restart event's `position` / `label` / `team`
  - `outcome` = scan the same log: `goal_within_30s` if a Goal in `[t, t+30s]`, else
[...]

Issue #44, opened 2026-07-11
[Task Proposal] Full-game NBA lineup state reconstruction


samliubearberkeley-oss:
## Task title

Full-game NBA lineup state reconstruction

## Family

`agentic_vbench_understanding`

## Task summary

Given a complete NBA game broadcast, reconstruct the exact five-player
on-court lineup for both teams at a deterministic series of official
game-state query points across the full game.

The intended query points are immediately after each atomic substitution
batch in the official play-by-play. Substitutions occurring at the same
period and game clock would be grouped and applied before the lineup
snapshot is emitted.

## Expected output

For every query point, the agent reports:

- query point ID
- period
- official game clock
- exact five-player away-team lineup
- exact five-player home-team lineup

Player identities would use a closed vocabulary of canonical IDs supplied
in the task prompt.

## Why this is long-horizon

This is a state-reconstruction task rather than independent substitution
detection.

To produce a correct lineup at a later query point, the agent must correctly
identify and accumulate every prior substitution. A missed or incorrect
substitution propagates into later lineup states.

The events are distributed throughout a complete game broadcast and should
require sustained full-game tracking and more than 50 tool-call turns.

## Ground truth

Ground truth would be generated deterministically from:

- official structured play-by-play
- official starters and roster data from the box score

The final task would pin the raw source files and use a reproducible state
machine to construct every lineup snapshot.

No manual video labeling would be required.

## Scoring

The primary reward would be exact snapshot accuracy.

A snapshot receives credit only when:

- the query point is correct
- both lineups contain exactly five unique valid player IDs
- the away lineup exactly matches ground truth
- the home lineup exactly matches ground truth

Order within each five-player set would not matter.

No partial credit would be awarded for four out of five players.

The verifier would be deterministic pure Python, with no LLM or VLM judge.

## Shortcut resistance

- A single frame cannot reveal the accumulated lineup state.
- A single substitution graphic only reveals one local event.
- Scoreboard OCR does not provide the complete five-player lineups.
- No-media and metadata-only attempts should score near zero.
- A less-famous game would be selected to reduce memorization risk.
- The evaluated agent would not receive the official play-by-play or box score.
- Internet lookup would not be permitted.

## Video plan

I plan to use an existing, complete, publicly accessible NBA broadcast rather
than collect new footage.

Before implementation, I will evaluate candidate games for:

- complete broadcast coverage
- stable downloadable video
- at least 720p resolution
- official structured play-by-play
[...]

PR #51, opened 2026-07-13
Add understanding task: soccer restart-outcome chains over a full match

Y-Yiqing:
Adds `tasks/agentic_vbench_understanding/soccer-restart-outcome-chains/`. Proposal issue: #50.

**Opening as a Draft.** The task is complete and passes `scripts/understanding/check_task.py` for structure and calibration. One design decision needs a maintainer call before it can build in Harbor: how to host an NDA-restricted broadcast. That question is at the bottom.

### The task

One full 90-minute broadcast, both halves concatenated (Bundesliga 2016-17, Mainz 05 1-1 Borussia Dortmund). For every visible ball restart the agent reports `(t, restart_type, team, outcome)`. Localizing a restart is the easy part. The scored fields are the taking team, read from kit colors held consistent across the match, and the outcome, which needs the 15 to 30 seconds of play after each restart to see whether a shot or goal follows. There are 81 visible restarts spread across the whole match.

Scoring is a grounded order-preserving 4-tuple F1 in pure stdlib (`steps/solve/tests/judge.py`). A predicted restart counts only if `restart_type`, `team`, and `outcome` all match a ground-truth restart and `|dt| <= 3 s`, one-to-one under an order-preserving alignment. No VLM or LLM judge.

Ground truth is a mechanical transform of SoccerNet-v2's published, multi-annotator `Labels-v2.json`, which is free and needs no NDA. `provenance/build_gt.py` derives it and asserts `verifier(oracle) == 1.0`.

### Calibration (measured)

|---|---|---|
| oracle | 1.0 | - |
| empty / null | 0.0 | - |
| no-media prior | 0.0247 | - |
| Claude Code (Opus 4.8) | 0.0225 | 110 |
| Codex | 0.0227 | 120 |
| Antigravity | 0.0449 | 120 |

All three strong agents land below 0.10 over more than 50 turns. Two independent agents (Claude and Codex) land within 0.0002 of each other, which is evidence the difficulty is real. Each agent found only 7 or 8 of the 81 restarts, so recall is the wall, and team and outcome errors compound on top. Rollouts are under `calibration/rollouts/`, one folder per agent, each with the final answer, the full `sampling_trace`, and a `requests.log`.

### Two honest caveats

1. The three calibration runs went through a frame-sampling tool, one frame per requested second under a hard 120-call budget, rather than the ffmpeg-on-baked-video harness the instruction describes. Every predicted `t` is an integer second and the turn count equals the number of frames sampled. The difficulty evidence is the same either way. A reviewer re-running in Harbor will use the baked video. This is noted in `calibration/scores.md`.
2. `no_media` and `random` ablations are measured near 0. The strong-VLM `single_frame` and `frame_dump_no_tools` ablations are argued in `SPEC.md` but not yet measured for this match. They will be added once the media is staged.

### Open question for reviewers: hosting an NDA-restricted broadcast

The one thing I want your call on before finishing the Dockerfile. The labels are public and free. The video is the wrinkle. SoccerNet distributes the full-match broadcast under an NDA that restricts public redistribution, and the copies are 224p or 720p rather than a pinnable archive.org or YouTube URL.

I looked at precedent in this repo:

- PR #6 migrated gated upstream materials onto public `agentic_vbench_*` HF datasets so anonymous workers can fetch them.
- The egocentric tasks #45 and #47 re-host an anonymized derivative of gated research footage on Hugging Face and bake from a pinned URL and SHA256.
- #44 raised the same "preferred video-hosting source" question for a broadcast task.

For an NDA-restricted broadcast, is the accepted path to re-host an anonymized derivative of the match, scoreboard and broadcaster logos blurred and re-encoded to at least 720p, on a public `agentic_vbench_understanding` HF dataset, then bake from a pinned URL and SHA256, the same pattern as #45 and #47? Or would you prefer a fully public broadcast source? The anonymization is something the task wants regardless, since identifying the match would let an agent look up the public labels.

Until this is settled, `environment/Dockerfile` carries a placeholder URL and fails fast with a clear message. Everything else runs today against a locally staged copy of the same match. Happy to go whichever way keeps this cleanly runnable and license-safe on your side.


christine1729:
Thanks for the PR. The verifier is careful work (deterministic, one-to-one matching, spam caps near zero) and the answer key stays verifier-side, exactly as it should. Feedback:

- [ ] State each run's model and use current ones (for example Fable 5, GPT 5.6 sol, Gemini 3.1 Pro or 3.5 Flash); the current rollouts don't name them.
- [ ] Include raw rollout transcripts, since the current summaries cannot be audited for tool use or web access.
- [ ] Measure the `single_frame` and `frame_dump_no_tools` ablations instead of arguing them.
- [ ] Fix the oracle check: `build_gt.py` has no `verifier(oracle)==1.0` assert, and `run_ablations.py` imports modules the PR does not include.
- [ ] State the 3-second tolerance in the prompt.

Smaller: the delivered match has 81 events rather than the 150+ from issue #50, so please confirm no public per-event timeline exists for it.

The scoring core is in good shape. Looking forward to the next version!

Y-Yiqing:
Thanks so much for the careful review, these were all fair catches. Everything is addressed in 541aff3:

- **Models**: every rollout now names its model (Fable 5 / Opus 4.8 / GPT-5.6-sol / Gemini 3.5 Flash), verifiable inside the transcripts.
- **Raw transcripts**: added for all runs under `calibration/rollouts/<agent>/transcripts/` (Codex CLI session logs, Antigravity's own SQLite conversation db, full Claude transcripts). Images elided, everything else verbatim; `rollouts/README.md` explains the formats.
- **Ablations measured**: single_frame = 0.0, frame_dump_no_tools = 0.023 (both Fable 5); artifacts in `provenance/ablations/measured/`.
- **Oracle check**: you were right, the assert was lost in a restructure. `build_gt.py` now asserts derived GT == `judge.py`'s answer key and oracle == 1.0; `run_ablations.py` is self-contained. Both run clean.
- **Tolerance**: now stated in `instruction.md`.

Two small extras: measuring the ablations surfaced a prompt wording mismatch on the outcome field (the GT counts a goal by either side; prompt now says so, GT and scores unchanged), and the transcript re-run added a fourth data point, Fable 5 at 0.0 over 118 turns.

christine1729:
Hi @Y-Yiqing, thanks for the rev.

Interesting model behaviors from this task!

techgenmini:
@Y-Yiqing — the Dockerfile still has the PLACEHOLDER `MATERIALS_URL`, so the task can't build; could you provide the video URL and its SHA256?

Y-Yiqing:
Thanks for building it, and sorry about the placeholder. The only full copy of that match is
SoccerNet's, which comes under an NDA that does not allow sharing it, so there was no URL we
could give. #121 moves the task to Dortmund 6-2 Leverkusen, which the Bundesliga has
published in full on YouTube, and re-measures the calibration in the built image.

Video: https://www.youtube.com/watch?v=U-Glif5abSY, format 136, baked as served
SHA256: 7648b622c196ca5c844eb6162dfda2532d8e46ced99728f720f62fb6d3d695e2

Calibrating in the image also showed the restart question was too easy for a strong agent,
so the same PR widens it to the match's full action ledger and explains why in
`calibration/scores.md`. It also removes 62 frames of the old broadcast that the merged
Antigravity transcript database had stored.

techgenmini:
@Y-Yiqing Thanks for fixing the video source. We ran the revised full action-ledger task in #121 with the exact video checksum, the native 3-hour budget, and web search disabled:
- Astra / Codex: 0.4880
- Opus 5 / Claude Code: 0.3737
- Gemini 3.8 Flash / Antigravity: 0.1352
All three completed normally. These are single-run results, but all three exceed 0.1, so the revised task still falls short of the difficulty target we’re aiming for.
We’re looking for tasks that expose substantial remaining gaps in video understanding. Could you iterate further on the task design and recalibrate? Ideally, the added difficulty should come from deeper temporal reasoning, cross-event dependencies, or evidence integration—not simply tighter timestamp tolerances, more annotations, or a shorter time limit.
We’ve preserved the trajectories, raw scores, submissions, and intermediate artifacts and can share them to help identify which parts the agents already handle well and where a more challenging task could focus.

Y-Yiqing:
@techgenmini Thanks for running #121 and for being specific about where the difficulty should come from. We took the brief seriously and built a new task rather than tightening the soccer one: **#124, `voxel-standing-block-census`**.

The change is what the answer *is*. In #121 it was a ledger of events, so an agent that finds most events scores well. Here each of the thirty answers is a **state** — how many rig-placed blocks of each type are standing at that moment — which is everything placed before it minus everything removed before it. One missed placement is wrong at that checkpoint and at every later one, and a count scores only when it is exactly right. There is no tolerance to tighten and no per-event partial credit. Removals are the hard half: a block that vanishes was placed earlier and is usually removed off camera, so it can only be inferred by comparing what is standing against what the agent believes it placed.

Calibrated with your three agents, each acting only inside the task container with `--network none`, and with the ground truth unreadable on the host for the whole of every run:

| agent | #121 (soccer) | #124 (this task) |
|---|---|---|
| Astra / Codex | 0.4880 | **0.0556** (104 turns) |
| Gemini 3.8 Flash / Antigravity | 0.1352 | **0.0056** (165 turns) |
| Opus 5 / Claude Code | 0.3737 | did not finish in the 4 h budget |

Oracle 1.0, best trivial submission 0.0222, the three degraded-input ablations 0.0056 / 0.0333 / 0.0056. `scripts/understanding/check_task.py` passes 9/9.

Two things we would rather flag than have you find: the Opus 5 arm spent the whole budget on a detection pipeline and never wrote an answer, so it scores zero by not finishing rather than by being wrong, and the PR says so; and the Codex figure is the higher of two runs (0.0389 on an earlier host-side run, 0.0556 in the container).

We would still like the preserved trajectories and intermediate artifacts you offered for #121 — they would tell us what the agents already handle well, which is useful whether or not the soccer task gets another pass. #121 stays as it is for now.

techgenmini:
@Y-Yiqing Thanks for the careful write-up. Building a new task was a reasonable call, and #124 looks like a solid start. Please go ahead and finish `voxel-standing-block-census`. Here are a few things we noticed while reading it. Please check them and adjust what makes sense to you:

1. **Where the difficulty comes from.** We compared each agent's per-minute net change per type (each checkpoint minus the previous one) against the truth. This takes out the cascade effect. Codex's deltas are exact in only 15.6% of cells (42.8% within ±1). Gemini's are exact in 3.9%. So even without the cascade, the agents miss a lot of events: at t=960, Codex counts 348 standing blocks against a true 454. Our reading is that the main bottleneck is dense change detection over a long recording, and the exact-cumulative metric makes that gap bigger. That is a real video-understanding gap and worth measuring. But it sits closer to `perception` than `understanding`. Could you check this yourself, and if you agree, say so in the SPEC?
2. **Are removals off camera?** The PR description and your #51 comment say removals are "usually off camera". SPEC §5 says all 315 removals are in frame. If some really are off camera, that is where the cross-event reasoning comes from, and it would be worth designing around on purpose. If they are all in frame, removal type can be read directly at the moment it happens. Either way, please make the two documents agree.
3. **Overlap with `minecraft-gameplay-ledger-s1`.** Both tasks are first-person voxel footage with block placement and removal identified by texture. It would help to explain in the SPEC what this task measures that the Minecraft one does not.
4. Some smaller items:
   - `scores.md` points to three `*-tool-calls.log` files that aren't in the PR, and there is no raw Gemini transcript.
   - Pinning the HF media URL to a commit instead of `main` would make builds reproducible.
   - `generator/run_arm.sh` and `generator/run_claude.sh` still contain local absolute paths.

After this is merged, we'll rerun it with our standard profiles, as we did for #121.

**Soccer (#121) trajectories.** Here is the preserved data from our three Modal runs: [this drive link.](https://drive.google.com/drive/folders/11q4E2k7on5HPSJVlQRJJNKe0OERWKdRc) (I shared it with your github gmail account yq725728@* let me know if any access issue)For each arm (Astra / Codex 0.4880, Opus 5 / Claude Code 0.3737, Gemini 3.8 Flash / Antigravity 0.1352) it includes:
- the raw and imported rewards
- the final submission
- the native CLI transcript and ATIF trajectory
- `/workspace/output` and `/workspace/work` intermediates
- per-arm SHA-256 manifests

One note: the Astra run used **medium** reasoning effort. If these runs point you to a way to make the soccer task harder, specifically through deeper temporal reasoning or cross-event dependencies, we'd be glad to see another iteration on it too.

Y-Yiqing:
@techgenmini Thank you, these were all worth checking, and on the first two you were right.

**1. Where the difficulty comes from.** Your numbers reproduce exactly: Codex's per-minute deltas are exact in 15.6% of cells (42.8% within ±1), Gemini's in 3.9%, and at t=960 Codex has 348 standing against a true 454. We split the error one step further. Codex's per-minute *total* is right in only 2 of the 30 minutes, but the *mix of types* it reports stays close to the truth (total variation distance 0.090). It names what it sees and misses what changes. We agree the bottleneck is dense change detection, and that the exact cumulative metric widens that gap rather than adding a reasoning step. `SPEC.md` now says so and sets `cognitive_level: perception`, with the justification spelled out in §1.

**2. Removals.** Our PR description and #51 comment were wrong, and the SPEC was right: all 315 removals are in frame when they happen (median apparent size 34 px, 28 px at the tenth percentile), so a removal's type can be read at that moment. We corrected the PR description. Given that, we kept the task as it is and described it honestly, rather than presenting it as cross-event reasoning.

**3. Overlap with `minecraft-gameplay-ledger-s1`.** `SPEC.md` has a new section with a side-by-side table. In short, that task makes each event easy to see and hard to name, and this one does the reverse. There, every action is framed dead-centre with the camera settled on it, a crack grows through each dig, the hotbar shows the held item, and the vocabulary has six logs and seven terracottas; events arrive at 8.4 per minute and the score falls with length (0.36 at 53 min, 0.020 at 238 min). Here, a block appears or vanishes between two frames with no cue while the camera keeps flying, placements are a median 24 px, the six types are easy to tell apart, events arrive at 39.3 per minute, and 30 minutes is enough for 0.0556. The type-mix result in (1) is what that separation looks like in the data.

**4. Smaller items.** All fixed:
- The three `*-tool-calls.log` files were in the task directory but silently dropped by the repository's `*.log` ignore rule. They are now `*-tool-calls.txt`. The same rule had also dropped the raw event log the truth is replayed from; it is now `generator/rig_events.tsv`.
- The raw Antigravity trajectory is added (`gemini-container.jsonl.gz`, 320 steps). It names no model, so `scores.md` records the identity from the conversation store: all 160 planner steps carry the identifier the app saved for the Gemini 3.8 Flash / medium selection.
- The media URL is pinned to commit `b026589`; Hugging Face's `x-linked-etag` for it equals the SHA256 in the Dockerfile. The oracle smoke re-fetched the video from that URL through the Dockerfile's SHA256 check and scored 1.0.
- `generator/` carries no machine-specific path anywhere, not only in the two runners; paths come from environment variables, and the lock directories are required so a calibration cannot run unlocked. While doing this we found `gt6_fixed.py` and `controls6b.py` imported two modules the PR did not include. They are added, and replaying the event log from a `git archive` of the branch now matches the shipped truth in all 180 cells and reproduces every attack in `scores.md`.

Thanks also for the soccer trajectories.

Issue #54, opened 2026-07-14
[Task Proposal] Volleyball ace and block timeline over a full college match


luojiaxuan:
# Task Proposal — Volleyball ace and block timeline from a full college match broadcast

**Family:** `agentic_vbench_understanding`
**Proposed task id:** `byu-wsu-2023-volleyball-ace-block-timeline`

## One-line

Given the full 115-min broadcast of #8 BYU at #13 Washington State women's volleyball (2023-09-08, WSU wins 3-1, four sets), find every point that ended as a **service ace** or a **block**, and report set, exact score right after the point, event type, and the credited player(s). Graded by strict F1 against the official NCAA play-by-play.

## Video

- Official WSU Athletics YouTube upload: https://www.youtube.com/watch?v=5KC1jC90lT8 (1080p available, 115 min, inside the 10-300 min window).
- archive.org mirror exists (`youtube-5KC1jC90lT8`, 480p); at build time we bake the >=720p original, re-host with pinned URL + SHA256.
- Audio track stripped at bake time (`ffmpeg -an`): pure video task, no commentary leak, audio-only ablation n/a by construction.

## Question and output

`/workspace/output/solution.json`:

```json
{"events": [
  {"set": 2, "score_after": "13-10", "type": "ace",   "players": ["First Last"]},
  {"set": 3, "score_after": "5-8",   "type": "block", "players": ["First Last", "First Last"]}
]}
```

- `score_after` is `"<BYU>-<WSU>"` right after the rally — BYU on the left, the same order the broadcast score graphic displays (verified from broadcast frames; the official record uses the same orientation). Scores only go up inside a set, so this anchors each event uniquely — it is this task's version of the game clock in the worked example, and it is readable only from the score bug at that rally.
- `players`: for an ace, the server; for a block, all credited blockers (one or two), order-insensitive.
- Both rosters (jersey number -> full name) are given in the instruction — college volleyball jerseys carry numbers, not names.

## Ground truth (machine-truth, zero hand labeling)

Official NCAA play-by-play (stats.ncaa.org), cross-checked with the WSU box score (wsucougars.com/sports/womens-volleyball/stats/2023/brigham-young/boxscore/18724) and BYU's own box score (byucougars.com/boxscore/20178). Volleyball PBP logs every rally with running score and credited players. Density: ~19 block points + ~8-10 aces = **~25-30 target events out of ~180 rallies**.

## Why hard and long-horizon

[...]

PR #59, opened 2026-07-16
Add task: byu-wsu-2023-volleyball-block-timeline (#54)

luojiaxuan:
Implements the task proposed in #54: reconstruct every service ace and block point across a full college volleyball match (#8 BYU at #13 Washington State, 2023-09-08) from the broadcast video alone.

### What's here
`tasks/agentic_vbench_understanding/byu-wsu-2023-volleyball-ace-block-timeline/`

- `steps/solve/instruction.md` — task + output schema (per event: `set`, `score_after`, `type` ace/block, `players`, and for blocks the `blocked` hitter) + rosters (college jerseys are numbers only, no names on the broadcast).
- `steps/solve/tests/judge.py` — deterministic pure-stdlib F1 grader: two-tier credit, greedy one-to-one multiset matching, no LLM and no network. Oracle → 1.0, empty/guess → 0.
- `steps/solve/solution/solve.sh` — the verified answer key: 24 events (5 aces, 19 block points) from the NCAA rally-by-rally log (contest 3241315), reconciled per player against both schools' box scores.
- `environment/Dockerfile` + `bake_media.sh` — bakes the match at build time from a **pinned URL + SHA256** (720p, audio stripped so the task is single-modality video). The pinned bytes are live and hash-verified during build.
- `calibration/scores.md` + `rollouts/` — full calibration with raw transcripts and sample block-moment frames.

### Reproduce
Build the image (it pulls and SHA256-verifies the video), then run `steps/solve/tests/test.sh` against any `solution.json`.

### Calibration (hardened judge, all runs > 50 tool-call turns)

| agent | model / setting | score | turns |
|---|---|---|---|
| Codex CLI | gpt-5.6-sol, xhigh | 0.02 | 126 |
| Claude Code | Opus 4.8 [1m], xhigh | 0.0 | 291 |
| Claude Code | Fable 5, xhigh | 0.15 | 733 |
| Antigravity | Gemini 3.5 Flash (High) | 0.0 | verified clean |

Four strong agents, all ≤ 0.15 (mean ≈ 0.06); oracle 1.0, empty/guess 0. The terminal ace/block moment is a sub-second net exchange and net-front jersey ID at 720p is the wall that keeps scores low. The Antigravity number required a three-vector isolation harness (always-on no-web rule + repo-external sandbox with `trustedWorkspaces` narrowed to the workspace + wiping agy's local memory); the full write-up and all transcripts, including the cheat-vector runs, are in `calibration/rollouts/` and #54.

Corresponds to the proposal in #54.


techgenmini:
Hi @luojiaxuan, thanks for the PR — scanning a full match for every ace and block is a strong test. The Fable/Opus/Codex spread is useful calibration evidence, but Fable’s `0.15` is an overage under the
  current rule, not an established variance allowance.

  - [ ] Remove the self-authored `<=0.109` reviewer allowance. The #54 gate question never received a maintainer ruling, and `0.15` exceeds either threshold.
  - [ ] Complete the SPEC card: difficulty and all ablations are still marked TBD/to be measured.
  - [ ] Measure actual no-media, single-frame, and frame-dump ablations. Empty/null is not a no-media run, especially because the NCAA PBP is public.
  - [ ] Re-run the final `blocked`-hitter schema in a clean task-image-equivalent workspace, with no prior-work artifacts. Record model, harness version, tool profile, turn count, and trajectory path; publish
  the full Claude/raw trajectories as a hash-pinned release/LFS/dataset asset rather than only digests.
  - [ ] Replace the two real target anchors in the instruction examples (`set 1, 7-7` ace and `set 2, 10-11` block) with synthetic examples.

  Happy to re-check a focused calibration update.

tasks/agentic_vbench_understanding/byu-wsu-2023-volleyball-ace-block-timeline/steps/solve/tests/judge.py
```diff
@@ -0,0 +1,276 @@
...
+    return isinstance(pred_list, list) and len(pred_list) == len(gt_list) \
+        and matched_names(pred_list, gt_list) == len(gt_list)
+
+
+def players_off_by_one(pred_list, gt_list):
+    """One name missing, one extra, or one wrong — everything else matched."""
+    if not isinstance(pred_list, list) or not pred_list:
+        return False
+    m = matched_names(pred_list, gt_list)
+    if m < 1 or m < len(gt_list) - 1:
```
techgenmini: The partial-credit description says one blocker may be missing, extra, or wrong, but a wrong or omitted **solo** blocker currently earns no partial credit (`m < 1` rejects it), while one wrong player in a
  two-blocker event can. Could you make the intended rule symmetric and add regression cases for solo-wrong, solo-missing, solo-extra, and one-wrong-pair?

tasks/agentic_vbench_understanding/byu-wsu-2023-volleyball-ace-block-timeline/steps/solve/tests/judge.py
```diff
@@ -0,0 +1,276 @@
...
+# blocker on the scoring team's opponent); the box score resolves the block point to
+# a Jehlarova solo, and the blocked hitter is unrecoverable from the corrupted line,
+# so `blocked` is None there and the judge does not require it for that one event.
+GROUND_TRUTH = [
+  {"set": 1, "score_after": "7-7",   "type": "ace",   "players": ["Erin Livingston"]},
+  {"set": 1, "score_after": "16-13", "type": "ace",   "players": ["Erin Livingston"]},
+  {"set": 1, "score_after": "18-14", "type": "block", "players": ["Whitney McEwan-Llarenas", "Elyse Stowell"], "blocked": "Katy Ryan"},
+  {"set": 1, "score_after": "18-17", "type": "block", "players": ["Argentina Ung", "Magda Jehlarova"],         "blocked": "Elyse Stowell"},
+  {"set": 1, "score_after": "20-17", "type": "ace",   "players": ["Aria McComber"]},
+  {"set": 2, "score_after": "1-2",   "type": "block", "players": ["Magda Jehlarova"],                          "blocked": None},
```
techgenmini: The instruction requires `blocked` for every block, but this GT row has `blocked: None` and the judge silently exempts that field. Could you recover the hitter from video/another auditable source, or exclude/
  restructure this event? A hidden exception makes the task contract and scorer disagree.

luojiaxuan:
Applied the fixes from your list that don't need fresh compute, plus one you flagged inline.

**Partial-credit symmetry (your inline note).** You were right: `players_off_by_one` counted a substitution as two errors, so one wrong name in a two-player block earned partial credit while a solo block credited to the wrong player earned nothing. Errors are now `max(unmatched-GT, unmatched-pred)`, so solo-wrong, solo-missing, solo-extra and one-wrong-pair all sit in the same tier, and two errors earn nothing either way.

`steps/solve/tests/test_judge.py` is new and pins exactly the four cases you named, plus the blocked-hitter tiers, the one event whose hitter the source lost, and the two-Bower lastname ambiguity. 12 cases, all passing; oracle still 1.0, empty still 0.0.

**This moved a published number, so flagging it explicitly:** re-grading the archived rollouts under the corrected scorer takes **Codex from 0.02 to 0.04**, because several of its solo-block credits were wrong rather than absent and now earn partial credit. Fable 0.15, Opus 4.8 0.0 and Antigravity 0.0 are unchanged. `scores.md` and `SPEC.md` carry the corrected figure and a section explaining the change.

**Real answer keys in the instruction.** Both examples (`set 1, 7-7` ace and `set 2, 10-11` block) were live entries in the key. They're now synthetic scores that appear nowhere in it.

**The `<= 0.109` allowance is gone**, everywhere. You're right that the gate question in #54 never got a maintainer ruling. For the record, it wasn't invented from nothing — it was pinned to a measured 0.106 re-grade recorded at the time — but it shouldn't have been written into the spec as an accepted threshold without that ruling, so it's out and the bar reads ~0.10.

Still open on this PR, and I'd rather not claim them until they're measured: the SPEC ablation fields, the actual no-media / single-frame / frame-dump runs, and a fresh run of the final schema in a clean workspace with full trajectories published.

The sister task #61 now has all of those done, including a forced-answer no-media ablation (the model submits a full-looking answer from recall of the public NCAA log and still scores 0.0) and an observability ledger that resolves whether the answer key is actually visible in the broadcast. Whatever you settle on there for trajectory publication and ablation protocol, I'll mirror here — worth agreeing the format once rather than twice.

luojiaxuan:
Worked the rest of the list, and found something while doing it that changed the task.

### The old numbers were measuring the wrong thing

You asked for a fresh run of the final `blocked`-hitter schema in a clean workspace. Doing that turned up why it mattered more than either of us assumed: **none of the four archived runs contained a `blocked` field at all.** They were produced against the pre-hardening instruction and then graded under the hardened scorer, so no matter how good they were they could never earn full credit on a single block. Codex 0.02, Fable 0.15, Opus 4.8 0.0 — all of them measured a schema the task had stopped shipping.

Fable's 0.15 in particular came entirely from aces, which need no `blocked` field. So the "Fable is strong at this task" reading was wrong; it was scoring on the one event class that didn't require the hardened attribution.

I've replaced those rows rather than amended them, and moved the old rollouts to `_superseded-old-schema/` with the reason recorded. The general lesson is in `scores.md`: after changing a task's schema, re-run the agents — do not re-grade their old answers.

### What the fresh run actually showed

On the real schema, in a clean workspace, **Codex scored 0.1887** — over the bar. Two causes:

**The aces were a giveaway.** One legible jersey read with the ball landing untouched. Dropped them; block-only.

**Two attributions weren't enough.** Block-only alone still left Codex at 0.1333, because with 19 events two correct blocks carry a lot of recall. So the scorer now asks for a third attribution: **the setter who fed the blocked attack**. That read sits mid-rally, before the attack, which is what makes it different in kind — the blocker and the hitter can both be recovered from the post-point close-up, but the setter cannot; you have to track the rally back. The official log records the whole chain (`Set by X → Attack by Y → Block by Z`), so the key is as solid as before.

### Calibration on the final schema

| agent | model / setting | score | work | integrity |
|---|---|---|---|---|
| Codex CLI | gpt-5.6-sol, xhigh | **0.0189** | 29 events submitted | key 0, web 0 |
| Claude Code | Opus 5, xhigh | **0.0** | 386 tool-call turns, 23 events | key 0, web 0 |
| Antigravity | — | not run | — | — |

Both wrote their own `solution.json`, both filled all three attributions on every event they submitted, and **neither got a single block point fully correct**. Between them they identified 2 of 18 rallies at all.

**Ablations** (Sonnet, forced to answer — a refusal measures nothing): no_media 16 events → 0.0, single_frame 9 → 0.0, frame_dump 16 → 0.0. no_media is the one that matters here since this match's log is public; forced to answer it produced 16 plausible events from recall and matched none.

### Your inline notes

**Solo partial credit** — fixed as you described, and the fix caught a second bug: the name-ambiguity table was computed over blockers and hitters only, so a setter appearing nowhere else in the key would have been matchable by lastname alone. It now covers setters too. `test_judge.py` is new: 17 cases covering each way one of the three attributions can be wrong versus two, solo/shared symmetry, the two-Bower ambiguity, anchors and duplicates. All passing.

**The `blocked: None` exemption** — gone, and not by special-casing. That event (set 2 at 1-2) is the corrupted rally line whose hitter and setter are both unrecoverable, so it is now excluded from the key outright. 18 events, every one fully specified, no hidden exceptions between the contract and the scorer.

**SPEC card** — all TBDs replaced with measurements.

### Package

Media pinned to an immutable dataset revision as well as its SHA256. Clean `--no-cache` build verified: media hash matches inside the image, **oracle 1.0 and null 0.0 run through the built image**.

One thing left that I'd rather you weigh in on: the directory and task ID still say `ace-block`, which is now wrong. I left it so this PR survives — renaming mid-review seemed worse than a stale slug. Happy to rename to `byu-wsu-2023-volleyball-block-timeline` before merge on your word.

techgenmini:
> Hi @luojiaxuan — this is a good and candid correction. I agree that the pre-hardening rows, including Fable 0.15, should be retired rather than treated as a gate exception. The scorer fixes, synthetic
  > examples, media pinning, and forced-answer ablations are real progress.
  >
  > I’m happy with the block-only direction; please rename this to byu-wsu-2023-volleyball-block-timeline before merge.
  >
  > A few final, bounded items:
  >
  > - [ ] Freeze one consistent 18-block contract: remove the remaining ace language in the instruction, and make SPEC/judge provenance say the corrupted set-2 1-2 source row is excluded rather than retained as
  >   an exemption. Replaying the committed Codex answer under the current judge gives 0.0213, not 0.0189; the no-media and single-frame reward files still report the old 24-event denominator.
  >
  > - [ ] Add a compact event-to-timestamp/crop ledger for the new setter field. The official PBP makes the GT credible, but we still need to establish that the added attribution is observable in the baked
  >   video.
  >
  > - [ ] Run one final frozen-task-image calibration each for Codex, Claude Code, and Antigravity (using the established three-vector isolation). Archive the native raw traces with immutable URL + SHA, image/
  >   harness/tool-profile details, and reconcile the claimed Claude 386 turns with the committed 140-turn histogram. No n=3 campaign is requested.
  >
  > - [ ] The 60-uniform-frame result is useful, but it is not the family’s all-frames-pasted/no-tools ablation; add that condition or request a scoped exception.
  >
  > Then I’m happy to recheck promptly.

tasks/agentic_vbench_understanding/byu-wsu-2023-volleyball-ace-block-timeline/task.toml
```diff
@@ -0,0 +1,39 @@
...
+build_timeout_sec = 2400.0
+cpus = 4
+memory_mb = 8192
+# A ~2h broadcast at 720p (video-only) is a couple of GB; give the build and trial room.
+storage_mb = 12288
+# The clip is baked at build time and the answer is not online: keep this false.
+allow_internet = false
+
+[environment.env]
+HF_TOKEN = "${HF_TOKEN:-}"
```
techgenmini: > HF_TOKEN is unnecessary for this public, build-baked asset and exposes a host credential to the task environment. Please remove this environment entry.

luojiaxuan:
Two corrections and the rename, all in `6733ac9`.

**The Codex score is 0.0213, not the 0.0189 I posted above.** The answer file is
unchanged — the grading was wrong. My harness pulls `judge.py` from the branch at
grading time, and it picked up the scorer before the key was cut to 18 events, so the
run was scored against a 24-event denominator. Regrading the committed answer with the
committed scorer gives 0.0213 (29 events submitted, 0 full, 1 partial). Opus stays 0.0,
the three ablations stay 0.0, oracle 1.0, null 0.0. The verdict does not move, but the
number in the table did, so here it is.

To stop that class of error, `steps/solve/tests/test_judge.py` now re-grades every
answer committed under `calibration/` and asserts the published reward. If the scorer
and `scores.md` ever disagree again, the tests fail.

Also corrected: the two agents located **3** of the 18 rallies between them, not 2, and
the tool-call histograms are regenerated from the raw transcripts (one was empty, one
counted only the first leg of a resumed session). Codex: 71 completed tool-call items.
Opus: 386 tool-call turns.

**Renamed to `byu-wsu-2023-volleyball-block-timeline`** — directory, task ID and PR
title. This was the open question from my last comment; rather than leave a task called
`ace-block` that scores neither aces nor two attributions, I have made the change and
it is trivially revertible if you would rather it merged under the old slug.

Two smaller things in the same commit:

- the exact prompt the calibration agents received is committed as
  `calibration/rollouts/instruction-as-run.md` and referenced by SHA256 from each
  `run-metadata.txt`, so a reviewer can check what was actually asked;
- the Antigravity trajectories moved to `calibration/agent-integrity/`, which is what
  they document — Gemini backends rebuild this match from the public rally log unless
  workspace isolation, sandbox narrowing and a memory wipe are applied together. They
  are evidence about agent behaviour on a match with a public answer, not calibration
  scores, and the directory now says so.

tasks/agentic_vbench_understanding/byu-wsu-2023-volleyball-ace-block-timeline/task.toml
luojiaxuan: Removed in `608ff89`. `[environment.env]` is gone from `task.toml` entirely; the media is public and baked into the image at build time, so the task environment never needs a host credential. Same removal applied to the sister USC task.

tasks/agentic_vbench_understanding/byu-wsu-2023-volleyball-ace-block-timeline/steps/solve/tests/judge.py
luojiaxuan: Fixed. `players_off_by_one` now counts name errors as `max(unmatched-key, unmatched-prediction)`, so a substitution is one error rather than two and the rule is symmetric: solo-wrong, solo-missing, solo-extra and one-wrong-pair all earn partial credit, two errors earn nothing either way. `steps/solve/tests/test_judge.py` pins exactly those four cases, plus the tiers for each of the three attributions being wrong versus two.

tasks/agentic_vbench_understanding/byu-wsu-2023-volleyball-ace-block-timeline/steps/solve/tests/judge.py
luojiaxuan: Excluded rather than exempted. That rally line (set 2 at 1-2) is corrupted in the raw log — its name field credits a blocker on the scoring team's opponent, and neither the stuffed hitter nor the setter is recoverable from it — so the point carries no answerable attribution and is out of the key. The key is 18 events, every one fully specified, and the judge has no special case. `SPEC.md` states the consequence: an agent that finds and reports that 19th block loses 1/19 of precision and nothing else.

luojiaxuan:
Two of your four items are closed, in `6733ac9`, `608ff89` and `dd73018`.

**1. The 18-block contract is frozen.** The ace language is out of the instruction —
`type` is `"block"` and nothing else, `players` describes blockers only, and the rules
list aces among the endings that are *not* entries. SPEC and the judge's provenance
comment both say the set 2 / 1-2 source row is **excluded from the key**, not exempted
inside the scorer, and spell out the consequence: an agent that finds and reports that
19th block loses 1/19 of precision and nothing else.

You are right about **0.0213**, and I had reached the same number independently just
before your comment. The answer file was never wrong — my harness pulls `judge.py`
from the branch at grading time and picked up the scorer from before the key was cut
to 18 events, so the run was graded against a 24-event denominator. `scores.md` and
`SPEC.md` now read 0.0213, the no_media and single_frame reward files are regenerated
against the 18-event key, and `test_judge.py` re-grades every committed answer and
asserts its published reward, so that class of drift fails the tests from now on.

**Turn reconciliation:** 386 = 140 + 246. The Opus run is two legs of one session — the
first was cut off by a network drop and resumed in place — and the committed histogram
had counted only the first leg. Both legs are published as separate files.

**4. The all-frames / no-tools ablation is measured.** The whole match goes into
context as 77 uniform stills and the shell is taken away: no seeking, cropping,
upscaling or scripting, only reading frames and writing an answer. It submitted 10
events, named every player `"BYU #unknown"`, and matched no rally anchor: **0.0**. That
is a separate row from `frame_dump`, which leaves the shell in place. Artifacts in
`calibration/ablations/`, pinned in `test_judge.py`.

**Raw traces are published**, contributor-owned and revision-pinned:
[`gavinlaw/agentic-vbench-calibration-trajectories`](https://huggingface.co/datasets/gavinlaw/agentic-vbench-calibration-trajectories)
at `2b90b57ba72e521de8bc0ed24c1d0470dafc5f95`. `scores.md` lists a
`resolve/<commit-SHA>` URL and whole-file SHA256 for both scored runs and all three
shell ablations, and `calibration/rollouts/run-envelope.md` records the CLI versions,
the tool profile each run was given, the workspace contents and the network envelope
those streams were audited against.

**Still open on this PR:**

- **the setter observability ledger.** I am building it verifier-side, the same way as
  the sister task: OCR the score bug to pin the second each of the 18 points is
  awarded, then sample back through the rally for the set. Two things about this
  broadcast made the first pass wrong and are worth recording: the file opens on the
  tail of a *different* match (Towson–WSU, the graphic reads MATCH POINT), and the bug
  carries a SET SCORE row that names the set outright, which is far more reliable than
  inferring set boundaries from where the score collapses. The pass is rerunning
  against that.
- **the frozen-task-image calibration and the Antigravity row.** On #61 you said you
  were not asking for a new Codex/Opus campaign there once the archived streams
  establish parity, and the envelope document above is my attempt to establish exactly
  that for this task too. If it does, I would rather add only the Antigravity row here
  than rerun Codex and Claude Code inside the image — but say the word if you want the
  in-image reruns and I will do them.

techgenmini:
> Hi @luojiaxuan — thank you. I rechecked dd73018 and the full review record. The archive work is real: the seven listed revision-pinned artifacts verify, the Codex 71 / Opus 386 counts reconcile, and I found
  > no evidence of actual web lookup in the retained command streams. The partial-credit fix, synthetic examples, rename, media pin, and HF_TOKEN removal are also closed.
  >
  > The remaining work is now bounded:
  >
  > - [ ] Resolve the 19-vs-18 contract. The instruction asks for every block, but the set-2 1-2 block is excluded and a correctly anchored report loses precision. Please either validate/include that row, or
  >   make such a report non-penalizing and add a regression test.
  >
  > - [ ] Complete the compact 18-event timestamp/crop ledger for both the setter and terminal attribution, or relax the setter field.
  > - [ ] Replace the final calibration evidence with one frozen-task-image run each for Codex, Claude Code, and Antigravity (or request a scoped Antigravity waiver). The current archive is useful but does not
  >   establish image parity: its claimed exact-prompt record differs from the prompt the raw trace actually reads; Codex uses host Tesseract/ImageMagick and Opus uses host libraries absent from the Docker
  >   image; and the envelope explicitly says the runs were not OS-network-isolated. This is a parity finding, not a lookup accusation. Please archive the replacement streams with hashes, image digest, minimal
  >   wrapper, tool profile, and task-action network isolation. No n=3 campaign is requested.
  >
  > - [ ] Withdraw and replace all_frames. Its submitted answer says that no frame was actually opened and verifier GT was readable; it also uses ace entries and omits setters, and no native trace is published
  >   for it. Please run a genuine all-frames-pasted/no-tools condition with accessible frames, verifier/GT isolated, the frozen block-only prompt, and a pinned raw trace.
  >
  > Opus 5 is acceptable here; I am not requesting an additional Fable 5 / Opus 4.8 row, a scorer rewrite, or a repeat of the three already archived shell ablations. Please also redact only the /Users/
  > luojiaxuan prefix in the retained integrity artifacts.

luojiaxuan:
First two of the four, `20ea589`.

**The 19-vs-18 contract is resolved the way you suggested — non-penalizing, with tests.** A prediction anchored at set 2 / 1-2 is now set aside by the scorer: it earns nothing and costs nothing, rather than being charged as a false positive. The reward JSON reports `n_predicted` (what the agent submitted) and `n_scored` (what the denominator uses) separately, plus `set_aside_unscored_anchor`, so the arithmetic is visible. Six regression cases pin it: reporting that point does not change the reward, it does not move the scored denominator, a wrong-name report of it is equally free, and submitting it alone is still 0.0. `SPEC.md` and `scores.md` say the same.

I kept it out of the key rather than validating it in, because the corruption is in the source row itself — the name field credits a blocker on the scoring team's opponent, and neither the stuffed hitter nor the setter survives — so there is nothing to validate against beyond the box score's "a Jehlarova solo". With it non-penalizing, an agent that finds all 19 is no longer punished for being right.

**Host paths redacted.** The `/Users/luojiaxuan` prefix is gone from the three retained Antigravity trajectories, replaced with `$HOME`. Nothing else in those artifacts changed.

Still working, in this order: the all_frames replacement, the frozen-image reruns, the setter ledger, Antigravity. On all_frames you were right, and it is worse than the answer alone shows — writing that up when the replacement run is done.

luojiaxuan:
**all_frames is withdrawn and replaced.** You were right, and the trace shows it is worse than the answer alone: the model never opened a single frame. Its own closing note says so — no shell, no `Glob`/`LS`, `Read` returning `EISDIR` on the directory, ~16 filename conventions guessed and none resolved — and it labelled the submission "a structurally plausible fabrication". I had withheld the shell and the directory tools together, which left it no way to *discover* the filenames. That measured my harness, not the task. The same failure was in the USC leg: 17 of 22 tool results were errors.

The replacement fixes the three things you named:

- **the frames are reachable** — every filename is listed in the prompt, so no directory listing is needed;
- **the verifier and key are isolated** — the workspace is `~/avb-allframes/<task>/`, outside the repository, so no judge, answer key or sibling run is on a path the model could open;
- **the prompt is the frozen block-only instruction**, unmodified except for the ablation clause, and committed verbatim as `calibration/ablations/all_frames.instruction-as-run.md`.

Tools left available: `Read` and `Write` only. `Bash`, `Glob`, `Grep`, `Edit`, `MultiEdit`, `NotebookEdit`, `Task`, `WebFetch` and `WebSearch` are all withheld.

Result — and this time the model executed the condition:

| task | frames opened | events submitted | anchors matched | score |
|---|---|---|---|---|
| USC–WSU | 80 of 80 (80 `Read` calls, 1 error) | 10 | 2 | **0.0** |
| BYU–WSU | 76 of 77 | 12 | 1 | **0.0** |

It is worth reading what the USC model did with the frames, because it is the right negative result rather than a failure to try: it reconstructed the whole set-score progression from the on-screen graphic, got the final 3-2 and every set score right, and then explained why the condition cannot yield block points — at a 90 s stride a rally is essentially never captured mid-action, and a single blurry instant near the net cannot distinguish a stuff from a kill through the block, a deflection kept in play, or an attack error.

**Native traces published**, same scheme as the rest, at revision `c66a1ed95a6664c1f423cc0b8f1eee4d2f242e01`:

| task | file | sha256 |
|---|---|---|
| USC–WSU | [`all_frames.stream-json.gz`](https://huggingface.co/datasets/gavinlaw/agentic-vbench-calibration-trajectories/resolve/c66a1ed95a6664c1f423cc0b8f1eee4d2f242e01/usc-wsu-2023-volleyball-block-timeline/ablations/all_frames.stream-json.gz) (21.0 MB) | `e6132b8b3a73fca3dd9205a6108ec24e6d8f6facf456b96b6ce494338a30a3be` |
| BYU–WSU | [`all_frames.stream-json.gz`](https://huggingface.co/datasets/gavinlaw/agentic-vbench-calibration-trajectories/resolve/c66a1ed95a6664c1f423cc0b8f1eee4d2f242e01/byu-wsu-2023-volleyball-block-timeline/ablations/all_frames.stream-json.gz) (17.0 MB) | `76afe71d06f59262629b8c00a71e069551ae3abd9dbf534a69d5261ec171d0c3` |

The exact prompt for each is in the same dataset directory. Both rewards are pinned in `test_judge.py` with the rest.

Next up is the environment-parity rerun; the revised image is built and the wrapper is written.

luojiaxuan:
Environment parity is done for this task, in `b9180b8`, and it moved the Codex number a long way — upward. Reporting it straight because it narrows the margin.

**Confinement.** Same construction as the sister task: the workspace the agent sees holds no video at all, the broadcast exists only inside a container built from `environment/Dockerfile`, and the only route to it is a wrapper that execs there. The container runs `--network none`. Recorded in `calibration/parity/environment.txt`:

```
image digest: sha256:9b4f1e096b3460f2921de1ba2da08d90df9314c5bf57103b85eee0042fe88de4
container network: none
media inside the image: ee887b18…8796dc60   (matches the pin)
ffmpeg, ffprobe, tesseract, convert, python3 + numpy 2.1.3, Pillow 11.0.0
(cv2 and scipy explicitly absent)
```

I took your second option — deliberately change the tool profile, then recalibrate — for the reason given on #61. If you would rather the profile stay lean, that is a rerun and I will do it.

**Results.**

| agent | in the task image | on the host |
|---|---|---|
| Codex CLI, gpt-5.6-sol xhigh | **0.0952** | 0.0213 |
| Claude Code, Opus 5 xhigh | **0.0** | 0.0 |

**0.0952 against a ~0.10 bar, and it is the first run on either task to get a block point fully correct**: set 1 at 18-14, with both credited blockers (McEwan-Llarenas, Stowell), the stuffed hitter (Ryan) and the setter (Ung) all right. Two partials besides. So the earlier line that no agent has ever fully solved an event on this task is no longer true, and `scores.md` and `SPEC.md` say so.

**Where the jump comes from matters, and it is not attribution.** Rally anchors found went 3 → 11; fully-correct events went 0 → 1. What the added tools buy is a scripted pass over the score bug — OCR plus template matching across a two-hour broadcast — which makes *finding* the rallies cheap. That is the tedious layer. The attribution layer moved by exactly one event. If you read the margin as too thin, the lean-image rerun is the lever, and it acts on the tedious layer rather than on the thing the task is testing.

Opus took 384 tool-call turns (231 Read, 152 Bash), submitted 8 events, matched one anchor and no attribution.

**Traces published** at revision `5719e3303f366594dd471a048a69ead60e4c2bcd`, with `environment.txt` and the exact prompt beside each. Every SHA256 in `scores.md` is checked against what the dataset actually serves — 10 of 10 match.

Still owed on this PR: the setter/terminal per-field ledger, and the note that the resumed leg of the archived host run was driven by a continuation prompt carrying the sister task's schema. Antigravity is still blocked on quota — its first window came back CLEAN after 973 frames but exhausted before committing an event, and a resume hit 429 immediately.

techgenmini:
Hi @luojiaxuan — thanks for reporting the upward score movement plainly and publishing the native streams. I checked the new parity artifacts. The richer image tool profile is reasonable: please keep
  Tesseract, ImageMagick, NumPy and Pillow. I am not asking for a lean-image rerun.

  One confinement boundary is still open. `./run` keeps access to and extraction from the original MP4 inside the container, but the bind-mounted derived frames are then processed directly by native host tools.
  The Codex stream runs substantive OCR/image scripts such as host `python3 work/analyze_scores.py`, and the Opus stream likewise uses host Python/Pillow and image processing outside `./run`. The host tool
  surface and filesystem are also visible to the agents.

  I found no evidence of source lookup; this is a task-image parity issue, not an integrity allegation. But it means the new `0.0952` / `0.0` rows cannot yet be described as “every task action inside the
  network-disabled shipped image.”

  Before spending more quota:

  - [ ] Route all substantive task processing—Python, OCR, transforms and montage generation—through the no-network task container. The host may carry model transport and display container-produced outputs, but
  should not provide task-processing tools.
  - [ ] Mark the current parity rows provisional and run one corrected Codex trial first. Please do not rerun Opus, Antigravity or the completed ablations yet. If the corrected Codex result lands at or above
  the gate, report it rather than tuning or repeating it; we can then make the smallest necessary call on the audited Opus zero and the Antigravity quota block.

  The setter/terminal ledger and continuation-prompt disclosure remain open as you already noted; no need to repeat that work in response to this comment.

luojiaxuan:
Corrected Codex trial is done, `80265ab`. **0.0426**, down from the 0.0952 the flawed harness produced. Reporting it as it landed.

**You were right about the boundary, and the fix goes further than a patched wrapper: the agent now runs inside the task image.** Not a wrapper around it — the CLI itself executes in a container built from `environment/Dockerfile`, so every command it issues necessarily uses that image's tools. There is no host tool surface for it to reach. A useful side effect: the prompt is now the shipped instruction *verbatim*, because inside the environment the video already sits where `workdir/setup.sh` puts it and nothing needs explaining about wrappers.

**Egress is default-deny.** `OUTPUT` policy DROP; the one hole is an allowlisting CONNECT proxy that takes the CLI's backend and refuses every other host by name. Verified before the run:

| target | result |
|---|---|
| `stats.ncaa.org` (the answer source) | refused by the proxy, logged |
| raw `1.1.1.1:443`, bypassing the proxy | `TimeoutError` — dropped by iptables |
| the model backend | allowed |

```
harness digest: sha256:32b717107eb2d3ba3db4607c0a7d73d034527395e60672f0c63e2f505081f9b7
task image under it: avb-byu-parity:latest
agent: codex-cli 0.151.0, inside the container
media inside: ee887b18…8796dc60  (matches the pin)
tools the agent has: ffmpeg, ffprobe, tesseract, convert, python3 + numpy 2.1.3, Pillow 11.0.0
                     (cv2, scipy absent)
```

**Result: 0.0426.** 29 events over 232 command executions and five hours; 9 of 18 rally anchors; **one block point still fully correct** — set 1 at 18-14, both blockers, the hitter and the setter.

| | old harness | corrected |
|---|---|---|
| reward | 0.0952 | **0.0426** |
| rally anchors | 11 / 18 | 9 / 18 |
| fully correct | 1 | 1 |

The drop is in *finding* rallies, not attributing them. Host tooling made a batch OCR pass over the score bug cheap; inside the image the same work costs more, so fewer anchors and lower precision. Attribution is unmoved at one event — which is the point of the task, and that event is still the only fully-correct one across both tasks.

**The egress log is published with the trace**: 1365 connections to the model backend, **zero to anything else**. That is a proxy's record rather than a grep over a transcript, and it is hashed in `scores.md` like the rollout.

**Two harness failures came first and neither produced a result**, so neither is reported as one:

1. a login shell reset `PATH`, so the CLI was never found — the run exited in under a minute with nothing;
2. my proxy left `create_connection`'s 30 s timeout on the upstream socket, so the model stream was torn down after 30 s of quiet while the agent ran a long command. That run reached `turn.failed` after five reconnects, with "write solution.json" still unstarted on its own todo list — though it had already found 18 qualifying rallies and said set 4 held eight of them, which is exactly right.

I mention the second because it would otherwise look like an unreported 0.0.

**Opus and the ablations are untouched**, as you asked. The Opus parity row is now marked *provisional* in `scores.md` and `SPEC.md` — measured under the flawed harness, not rerun. Antigravity is unchanged. Ready for whatever call you want to make on those two given a corrected gate-setting row that lands at 0.0426.

The setter/terminal ledger and the continuation-prompt disclosure are still open; `calibration/observability/` now has all 18 flip times confirmed and a `setter-chain.md`, and I will finish the terminal per-field pass separately.

luojiaxuan:
The terminal ledger is in, at `calibration/observability/witness.md` — one row per event
per field, saying whether each credited blocker and the blocked hitter is legible in the
shipped video, with the eight-frame strip behind every row under `witness/`.

Result: every credited blocker legible on 8 of 18 events, one of the pair on 2 more, none
on 8. The blocked hitter reads on 4 of 18. Where a number is legible it agrees with the
key — no event where the video contradicts the official record.

Two things the strips turned up that are worth knowing if you check any of them:

- the cut to the post-point close-up is not at a fixed offset from the score bar, so the
  window spans both sides of it (t-7.0 to +3.2). On set 3 at 14-18 the close-up is over
  before the bar updates; on set 4 at 16-23 it comes after. A one-sided window would score
  several of these "not legible" wrongly.
- set 3 at 14-18 is the trap case: the close-up right after the play is of BYU 14, who is
  not credited on it.

This is the same shape as the USC ledger on #61, so the two are comparable. That was the
last item I had open on this PR.

Issue #53, opened 2026-07-14
[Task Proposal] Baseball run-scoring timeline over a full college game


luojiaxuan:
# Task Proposal — Baseball run-scoring timeline from a full college game broadcast

**Family:** `agentic_vbench_understanding`
**Proposed task id:** `asu-wsu-2024-baseball-run-scoring-timeline`

## One-line

Given the full broadcast of a normal college baseball game (Arizona State at Washington State, 2024-03-22, 192 min, final 8-7, 15 runs total), reconstruct every run: inning, half, **which runner scored**, the batter at the plate, and how the run scored. Graded by strict F1 against the official NCAA play-by-play.

## Video

- Official WSU Athletics YouTube upload: https://www.youtube.com/watch?v=s-VS4Z1hEaA (1080p60 available, 192 min, inside the 10-300 min window).
- An archive.org mirror already exists (`youtube-s-VS4Z1hEaA`) but only 480p, so at build time we bake the >=720p original and re-host with pinned URL + SHA256, same as the worked example's Dockerfile suggests.
- We strip the audio track at bake time (`ffmpeg -an`). So it is a pure video task: commentary cannot leak attribution, and audio-only ablation is n/a by construction.

## Question and output

`/workspace/output/solution.json`:

```json
{"runs": [
  {"inning": 3, "half": "top", "scorer": "First Last", "batter": "First Last", "event": "home_run"}
]}
```

- `event` closed vocabulary: `single, double, triple, home_run, sacrifice_fly, sacrifice_bunt, groundout, flyout, fielders_choice, error, wild_pitch, passed_ball, walk, hit_by_pitch, balk, stolen_base, double_play, other`.
- Both team rosters (jersey number -> full name) are given in the instruction, because college jerseys mostly only show numbers. Roster is input metadata, not an answer leak.

## Ground truth (machine-truth, zero hand labeling)

Official NCAA play-by-play (NCAA.com game 6235669 / stats.ncaa.org), cross-checked with the WSU box score (wsucougars.com/sports/baseball/stats/2024/arizona-state/boxscore/19272) and ASU's own box score — three independent official records. The PBP states exactly who scored, during whose at-bat, on what event.

## Why hard and long-horizon

[...]

PR #61, opened 2026-07-18
Add task: usc-wsu-2023-volleyball-block-timeline (#53)

luojiaxuan:
Second understanding task, replacing the retired baseball proposal (#53): reconstruct every **block point** across a full college volleyball match (#25 USC at #9 Washington State, 2023-11-12, five sets) from the broadcast video alone. Sibling of #54 but block-only and a different match, so it is a genuinely distinct task (see "Why block-only").

### What's here
`tasks/agentic_vbench_understanding/usc-wsu-2023-volleyball-ace-block-timeline/`

- `steps/solve/instruction.md` — task + schema (per block: `set`, `score_after`, `type`, `players` = credited blocker(s), `blocked` = the stuffed hitter) + rosters (college jerseys are numbers only).
- `steps/solve/tests/judge.py` — deterministic pure-stdlib F1 grader: two-tier credit, greedy one-to-one multiset matching, no LLM and no network. Oracle → 1.0, empty/guess → 0.
- `steps/solve/solution/solve.sh` — the verified answer key: 23 block points (USC 10 / WSU 13) from the NCAA rally-by-rally log (contest 3252428), reconciled per player against the box score (Block Solos 3/3 + Block Assists 14/20).
- `environment/Dockerfile` + `bake_media.sh` — bakes the match at build time from a **pinned URL + SHA256** (720p, audio stripped so the task is single-modality video).
- `calibration/scores.md` + `rollouts/` + `frames/` — calibration with genuine agent outputs and block-moment frames (jerseys + score bug legible at 720p).

### Why block-only
Built first as ace+block like #54, but calibration exposed the 5 service aces as a score leak: Fable nailed 4 of the 5 aces (a legible single-jersey read with the ball landing untouched) and, reporting few but precise events, hit **0.24** — over the bar — purely off aces while getting 0 blocks right. Dropping the aces removes the only legible event class; every target is now a two-jersey net read (the blocker multiset AND the stuffed hitter), neither of which is ever on the score bug. That also makes it distinct from #54.

### Calibration (block-only)

| agent | model / setting | score | notes |
|---|---|---|---|
| Codex CLI | gpt-5.6-sol, xhigh | 0.015 | 43-event genuine run, 0 blocks fully correct |
| Claude Code | Fable 5, xhigh | 0.031 | ~597 tool-call turns, 0 blocks fully correct |
| Antigravity | — | not re-run here | #54's 3-vector isolation recipe applies unchanged |

Both strong agents got 0 blocks fully correct; both genuinely sampled the video (hundreds of ffmpeg frame reads, well over 50 tool-call turns). Oracle 1.0, empty 0. The two numbers are block-only re-grades of the agents' complete runs — fresh block-only re-runs kept getting cut off by run-environment issues (a session quota limit, then a network drop), never the task, and block attribution was 0 full matches in every attempt (`calibration/scores.md` documents this). On Antigravity: the recipe from #54 (no-web AGENTS.md rule + off-repo workspace + wiped memory stores) transfers directly, and block-only makes web grounding the dominant risk since the one derived field (the blocked hitter) is named in the public play-by-play.

### Reproduce
Build the image (it pulls and SHA256-verifies the video), then run `steps/solve/tests/test.sh` against any `solution.json`.

Background: #53.


techgenmini:
Hi @luojiaxuan — the #53 update made an evidence-based pivot: you retired the baseball task when its true score floor was too high, then removed aces once the ace+block calibration showed a real shortcut. The
  block-only task has a solid verifier-side GT, an audio-stripped input, and a sensible attempt at partial credit.

  Because this volleyball replacement first appeared in the #53 follow-up rather than a separately reviewed proposal, I think it needs one focused validation pass before we treat the calibration as final.

  **Task and scorer validation**

  - [ ] The cited official PBP and box score do support the 23 GT entries. Please make that derivation durable and reviewable with a verifier-side source snapshot/hash and per-event source-row mapping or
  builder, rather than relying on duplicated hard-coded literals. Please also correct the prose to describe the PBP’s single `Attack error by X (block by Y …)` record.
  - [ ] The four 720p examples are useful, but the scorer requires exact one/two-blocker credit and the blocked hitter for all 23 rows. Add a calibration-only observability ledger for every event: video
  timestamp/window, before/contact/after score evidence, and confirmation that each required jersey is recoverable. Include a few short strips for the difficult cases; revise any field that cannot be
  substantiated from video.
  - [ ] Fix the partial-credit edge case noted inline, then add regression tests. Also remove the now-unreachable ace aliases/branches from `judge.py` so the scorer is plainly block-only.

  **Fresh calibration**

  - [ ] Keep the current ace+block re-grades as useful exploratory evidence, but run the final block-only instruction fresh on the required current-model matrix. The old prompt, search strategy, and false-
  positive denominator differ from the final task.
  - [ ] Mark Antigravity as not run until there is a per-task rollout, rather than reporting a sister-task recipe as a score.
  - [ ] Measure no-media, single/sparse-frame, and frame-dump ablations under the same declared runner policy, and retain complete raw trajectories for every scored run. The current answer files and tool
  histogram are not enough to audit prompt parity or the no-public-PBP-lookup rule.

  **Package cleanup**

  - [ ] Rename the task to a self-describing block-only ID before merge; the family README requires this, and IDs become stable after merge.
  - [ ] Update the Hugging Face dataset README to document this exact USC–WSU asset, its source, transformation, and rights basis; pin the dataset revision as well as the content SHA.
  - [ ] Rebase onto current `main`, clean-build the image, and attach oracle/null results from that built image.

  A small fresh pilot covering the observability ledger, scorer fix, and ablations would be an excellent next revision; then the full model matrix will be meaningful.

  Inline on players_off_by_one():

  The partial-credit description says a blocker may be missing, extra, or wrong, but a wrong or omitted solo blocker currently receives no partial credit while a one-wrong two-player block does. Could you
  define the intended solo behavior and add tests for solo-wrong, solo-missing, solo-extra, and one-wrong-pair cases?

luojiaxuan:
Worked through the list. Everything below is measured, and the artifacts are in the branch.

### Task and scorer validation

**GT derivation, durable and reviewable.** `calibration/gt-source/` now carries a canonical snapshot of the official rally-by-rally log (every `[visitor, score, home]` row of the five per-set tables; the site 403s non-browser clients, so it was captured in a browser with an end-to-end sha256 recorded both in the capturing page and after transfer), plus `build_ground_truth.py`, which rebuilds the 23 events from that snapshot with a per-event source-row mapping and asserts equality with `judge.GROUND_TRUTH` — verified 23/23. On the prose: there is no `Attack error by X (block by Y ...)` record in this contest's tables (0 occurrences, checked case-insensitively across every cell), so I did not add that description; the snapshot lets you confirm it directly.

**Observability ledger — this was the important one.** `calibration/observability/` has, for all 23 points, the exact second the point is awarded (derived verifier-side by OCR'ing the score bug: coarse pass, bisection, read-back confirmation — no agent output involved) and four frames per event spanning the post-point window.

What it shows about this broadcast: the wide sideline shot never resolves jersey numbers, and there are no replays; but a net close-up 0.8–3.2 s after each point does resolve them, and that window is the only place attribution is recoverable. On every event examined across the full window the credited blockers were legible and matched the answer key exactly (set 1 4-2 reads 91 and 10; set 2 16-10 reads 15 and 12; set 4 10-8 reads 8 and 21). The blocked hitter is legible less often — the close-up follows the net, which usually favours the side that won the point — which is why the scorer gives partial credit when the hitter is wrong but the blockers are exact. No answer-key name turned out to be unsupported by the video.

This also explains the calibration numbers: the Opus run concluded blockers "face away from the camera and are never readable", but it sampled the rally rather than that window. The information is narrowly placed, not absent.

**Scorer fix + regression tests.** `players_off_by_one` now counts errors as `max(unmatched-GT, unmatched-pred)`, so a substitution is one error and the rule is symmetric: solo-wrong, solo-missing, solo-extra and one-wrong-pair all earn partial; two errors earn nothing. `steps/solve/tests/test_judge.py` covers exactly those four cases plus the hitter tiers, anchor mismatches and duplicates — 17 cases, all passing. The unreachable ace aliases and branches and the `blocked: None` exemption are gone; the scorer is plainly block-only.

### Fresh calibration

Fresh runs on the final block-only instruction, each in a workspace holding only the video and the instruction with no prior artifact of this task reachable. `run-metadata.txt` beside each rollout records CLI version, model, effort, instruction hash, media hash and judge commit.

| agent | model / setting | score | turns | integrity |
|---|---|---|---|---|
| Codex CLI | gpt-5.6-sol, xhigh | **0.0185** | 160 items / 289 commands | key 0, web 0 |
| Claude Code | Opus 5, xhigh | **0.0** | 270 | key 0, web 0 |
| Claude Code | Fable 5, xhigh | not scored | 210 before that model's credits ran out | key 0, web 0 |
| Antigravity | — | **not run** | — | — |

Both scored runs ended normally and wrote their own `solution.json`. Neither got a single block point fully correct. Opus's closing report is archived verbatim and is the clearest statement of the task's shape I have: it rebuilt the entire 200-rally timeline from the score bug — reconciling exactly with the final box score, including two overturned challenges — and then said it could confirm exactly one block point, and that even there the blocker credit was "an inference, not a reading". The timeline layer is tractable; the attribution layer is not.

Fable is archived unscored rather than reported as 0.0: 210 turns of genuine frame work, then that model's credit pool ran out before it wrote an answer. Antigravity is marked not run, per your point — no number until there's a per-task rollout.

**Ablations, forced to answer** (a refusal measures nothing, so each demands a best-effort answer):

| ablation | inputs | events submitted | score |
|---|---|---|---|
| no_media | instruction only | 15 | 0.0 |
| single_frame | one midpoint frame | 11 | 0.0 |
| frame_dump | 60 uniform frames, no seeking | 22 | 0.0 |

`no_media` is the one you singled out, since this match's log is public: forced to answer, the model produced 15 plausible-looking events and matched none. The per-event score anchors and blocker/hitter pairs are not recallable.

### Package

- **Renamed** to `usc-wsu-2023-volleyball-block-timeline` (the branch name is unchanged so this PR survives).
- **Instruction examples** now use synthetic non-GT anchors (set 2 5-9, set 4 21-18) instead of real events.
- **Media pinned to an immutable dataset revision** as well as its SHA256, so the URL cannot start serving other bytes if the dataset moves. The dataset README documents this asset's provenance, exact transformation (`yt-dlp` 720p + `ffmpeg -an -c:v copy`, no re-encode), rights basis and takedown path.
- **Rebased onto current main** (clean, 7 commits).
- **Clean `--no-cache` build verified**: media hash inside the image matches the pin, oracle **1.0** and null **0.0** run through the built image.
- The unsupported `<= 0.109` language is gone everywhere; the bar is ~0.10.

One thing I did not do: raw stream logs run to hundreds of MB per run, so `rollouts/` carries each run's answer file, provenance and tool-call histogram rather than the raw streams. If you want the full streams as a hash-pinned dataset asset, say so and I'll upload them alongside the media.

tasks/agentic_vbench_understanding/usc-wsu-2023-volleyball-block-timeline/calibration/scores.md
```diff
@@ -0,0 +1,96 @@
...
+must be 1.0 and an empty attempt near 0.
+
+## Invariants, from the built image
+
+Built clean (`docker build --no-cache`) from `environment/Dockerfile`, which pulls
+the media at a pinned dataset revision and verifies its SHA256 during the build:
+
+| run | reward |
+|---|---|
+| oracle (`steps/solve/solution/solve.sh`) | **1.0** |
```
techgenmini, line 15: > Thanks — the source snapshot/builder and scorer regression coverage resolve my prior GT and solo-partial-credit comments. One calibration item remains: this line still marks the fresh final block-only run as pending, while the reported agent rows are re-grades of the earlier ace+block task. Please replace it with completed calibration under the shipped block-only instruction and retain the scored output and trajectory, so we can recheck task parity.

luojiaxuan:
One addition since the last comment, and a note on the matrix.

**The model matrix is complete as reported**: Codex CLI (gpt-5.6-sol, xhigh) 0.0185 and Claude Code (Opus 5, xhigh) 0.0, both fresh on the final block-only instruction in clean workspaces, both ending normally and writing their own answer, neither getting a single block point fully correct.

The Fable 5 run stays archived and unscored, and I want to be explicit that this is a decision rather than a gap. It reached 210 tool-call turns of genuine frame work before that model's credit pool ran out. The convenient move would have been to finish it with another model and report the result under Fable's name, so I tested that instead of assuming: resuming that exact session with Opus 5 scored **0.0**, and the three block points it submitted matched no rally at all — not the right rallies with wrong names, the wrong rallies. Inheriting 210 turns of another model's exploration changed nothing. Written up in `calibration/rollouts/hybrid-fable-then-opus.md`.

That experiment also produced the strongest evidence I have for the task's difficulty being real. **Both models independently reached the same wrong conclusion**: that jersey numbers at the moment of contact are unreadable (the hybrid's closing note says "~10 px tall and motion-blurred, and 5x magnification did not recover them"). They are right about that frame and wrong about the task — the numbers are legible, in the post-point close-up rather than at contact. Two strong agents failing the same way, for a reason the observability ledger shows is avoidable, is what I would want to see before believing a low score reflects difficulty rather than impossibility.

A method note in case it saves someone else the debugging: `claude --resume` restores the session's recorded working directory, so copying a session file into a second project directory isolates the conversation but not the filesystem — the resumed agent still writes into the original workspace. Isolating a resumed run needs both.

techgenmini:
Hi @luojiaxuan — this is a strong revision. I rebuilt the GT from the archived source snapshot (23/23), replayed the retained answers with the current judge (Codex `0.0185`, Opus `0.0`, all three ablations
  `0.0`), and confirmed the final instruction hash matches the run metadata. I’m not asking for a new Codex, Opus, or ablation campaign.

  Two final calibration-evidence items remain:

  - [ ] Please add a per-task Antigravity rollout under the documented no-lookup isolation, with its answer, score, and raw trace. Marking it “not run” is the correct reporting fix, but it does not complete the
  required three-agent matrix. If that backend cannot be made auditable, please request a scoped maintainer waiver rather than relying on the sister-task result.
  - [ ] Archive the native raw streams for every scored final/ablation run as immutable hash-pinned release or HF assets, and add each URL plus whole-file SHA256 to `scores.md`. The current `rollouts/README.md`
  says raw streams are omitted and the fresh histogram files are empty, so the turn counts and no-web/key-access audits are not independently checkable yet.

  I left two small evidence/package comments inline.

tasks/agentic_vbench_understanding/usc-wsu-2023-volleyball-block-timeline/calibration/observability/README.md
```diff
@@ -0,0 +1,81 @@
...
+| set 1, 25-15 | K. Williams, Ariail | opposing side in shot; not legible at this instant |
+| set 2, 10-7 | Ariail | still the wide shot at this instant; not legible |
+
+The last two illustrate the point above: a single sample from the window is not
+enough, which is why the sheets carry four frames each and why an agent that samples
+only the rally — as the Opus 5 run did — comes away with nothing.
+
+## Findings
+
+- **Credited blockers are recoverable.** On every event examined with the full
```
techgenmini: > These sheets are useful, but they do not yet demonstrate every scorer-required name. For example, set 1 / 14-10 says the blocker is not in the cited sheet, and this section says the blocked hitter is
  > visible less often, while full credit requires both fields. Please add a timestamp/frame/crop witness for every required blocker and blocked hitter, or relax any field that cannot be shown recoverable. This
  > is an evidence clarification, not a claim that the task is impossible.

tasks/agentic_vbench_understanding/usc-wsu-2023-volleyball-block-timeline/environment/Dockerfile
```diff
@@ -0,0 +1,34 @@
...
+#     https://www.youtube.com/watch?v=RgJjVgi7rZY
+#     "WSU Volleyball: #25 USC at #9 Washington State | Full Match | 11/12/23"
+#   Prepared: downloaded at 720p, audio track stripped (ffmpeg -an) so the task is
+#     single-modality video, then re-hosted at a stable URL so the bytes are pinned.
+#
+# Baked 2026-07-15 via environment/bake_media.sh (yt-dlp 720p + ffmpeg -an) and
+# uploaded to the dataset repo below; the build verifies the SHA256.
+# Pinned to an immutable dataset revision, so the URL cannot start serving other
+# bytes if the dataset's main branch moves; the SHA256 below is checked regardless.
+ARG MATERIALS_REVISION=859cb6877dc31b75d336648c4d3c87509e3373ab
```
techgenmini: > The media bytes are pinned correctly, but revision 859cb… predates the USC asset’s dataset documentation: its README lists only the ASU and BYU files. The README is now complete at HF revision
  > 7dc08d8a25c1192b411f586e60db6542776ae075, with the same USC file SHA. Please bump this revision and the matching SPEC.md URL/revision together; no media or calibration rerun is needed.

luojiaxuan:
Final pass on this one, `494a81d`. No score changed — I re-derived all of them from the
committed artifacts and they reproduce: Codex 0.0185, Opus 0.0, the three ablations
0.0, oracle 1.0, null 0.0. What changed is the evidence behind them.

**The published rewards are now a test.** `steps/solve/tests/test_judge.py` re-grades
every answer under `calibration/` and asserts the number in `scores.md`. If the scorer
and the write-up ever drift apart, the tests fail rather than the table quietly going
stale.

**The tool-call histograms are regenerated from the raw transcripts.** Three of the
five files were empty and one counted only part of a resumed session, so the counts in
the write-up had no committed source. They do now: Codex 158 completed tool-call items
(144 shell commands, 11 file writes, 3 MCP calls), Opus 270 tool-call turns, Fable 210.
Each file states which unit it counts, because the two CLIs do not log the same thing.

**One thing the Codex transcript shows that I had not reported.** Three of its tool
calls are attempts to leave the workspace: it listed the host's applications, then
asked for QuickTime Player and for a browser runtime. The host refused both
("Computer Use was not approved", "No browser is available"), the only URL string
anywhere in the transcript is `https://example.com/`, and no external content reached
the run — so the 0.0185 stands. But an agent reaching for a browser on a task whose
answer is public is worth stating rather than summarising as "web 0", and
`scores.md` now says it.

Also: `codex-aceblock-run.solution.json` is gone (nothing in the final write-up cites
it), `fable-run.solution.json` is now described for what it is — the partial event list
recoverable from the interrupted transcript, not something the agent submitted — and my
own session decision log is out of the task contribution.

Still yours to call: whether you want the full raw trajectory streams uploaded as a
hash-pinned dataset asset. They are hundreds of MB, which is why only the answers,
prompts, metadata and histograms are in git.

techgenmini:
> Hi @luojiaxuan — thank you for the thorough final pass. I checked the current revision, reran the new regression test, rebuilt the 23 GT rows, and reproduced the published Codex 0.0185, Opus 0.0, and all
  > three ablations at 0.0. The refreshed histograms and disclosure of the refused host-app/browser attempts are helpful.
  >
  > This closes the answer-to-judge consistency question. A few previously raised evidence items remain before approval:
  >
  > - [ ] Please publish the native raw streams for the two scored final runs (Codex and Opus) and the three scored ablations. They do not need to live in git: a revision-pinned Hugging Face dataset archive is
  >   appropriate (a contributor-owned dataset is fine if there is no project-owned archive). Record a resolve/<commit-SHA> URL and whole-file SHA256 for each stream in scores.md, plus the corresponding per-run
  >   metadata.
  >
  > - [ ] Please add one final Antigravity rollout under the documented no-lookup isolation, with its answer, score, and raw trace. This is not a request for an n=3 campaign. If you prefer not to run it, please
  >   explicitly request a scoped maintainer waiver for the missing third-agent row.
  >
  > - [ ] With the archived runs, record the actual image/harness version, tool profile, and no-network execution envelope. If the existing streams establish this parity, no Codex/Opus/ablation rerun is needed;
  >   rerun only an affected row if they do not.
  >
  > The two small inline items below also remain. Once these are addressed, this should be ready to move forward.

tasks/agentic_vbench_understanding/usc-wsu-2023-volleyball-block-timeline/calibration/observability/flips.json
techgenmini, line 1: > The sheets are useful, but this ledger still repeats GT and score timing rather than mapping every required blocker and blocked hitter to a concrete timestamp, frame, or crop. Please add those per-field
  > witnesses, or relax any field that cannot be shown recoverable from the supplied material.

tasks/agentic_vbench_understanding/usc-wsu-2023-volleyball-block-timeline/environment/Dockerfile
techgenmini: > Please bump this and the matching SPEC URL to 7dc08d8a25c1192b411f586e60db6542776ae075. The video SHA is unchanged, so this is provenance cleanup only and does not require recalibration.

luojiaxuan: Done in `ee66a0c` (the bump itself landed in `26767b7`). `environment/Dockerfile` and the `SPEC.md` url/dataset_revision now both read `7dc08d8a25c1192b411f586e60db6542776ae075`. I checked the two revisions serve identical bytes before switching — the USC blob is `13ccbabb…` at both — so no rebuild or recalibration was involved.

tasks/agentic_vbench_understanding/usc-wsu-2023-volleyball-block-timeline/calibration/scores.md
luojiaxuan: Resolved: the rows in the table are now completed runs on the shipped block-only instruction, not re-grades. Codex CLI (gpt-5.6-sol, xhigh) 0.0185 and Claude Code (Opus 5, xhigh) 0.0, each in a workspace holding only the video and the instruction. Answers, metadata and tool-call histograms are in `calibration/rollouts/`, and the native raw streams are now published — see the main thread.

luojiaxuan:
Two of the three items are done; they landed a few minutes after your comment, so they
were not in what you read.

**Raw streams are published.** Contributor-owned, revision-pinned, public:
[`gavinlaw/agentic-vbench-calibration-trajectories`](https://huggingface.co/datasets/gavinlaw/agentic-vbench-calibration-trajectories)
at `2b90b57ba72e521de8bc0ed24c1d0470dafc5f95`. `scores.md` now carries a
`resolve/<commit-SHA>` URL and whole-file SHA256 for each of the five scored runs
here — codex-fresh, opus-fresh and the three ablations — plus the Fable run that is
archived unscored. Each archive is the CLI's own output, gzipped and otherwise
untouched; the dataset's `MANIFEST.json` repeats the SHA256 of the `.gz` and adds the
SHA256 of the uncompressed stream inside it, so either can be checked:

```
sha256sum <file>.gz                  # matches sha256_gz
gunzip -c <file>.gz | sha256sum      # matches sha256_uncompressed
```

I verified all 14 blobs against the local manifest after upload rather than trusting
the uploader's exit code.

**Execution envelope recorded** — `calibration/rollouts/run-envelope.md`. CLI versions
(codex-cli 0.144.4, Claude Code 2.1.248/2.1.202), model and effort per run, the tool
profile each one was given, the workspace contents, and the network envelope. The
short version: the runs are not OS-network-isolated, since each CLI needs its own
model backend; isolation is at the tool layer (WebFetch/WebSearch/Task withheld where
the CLI supports it) plus an after-the-fact grep of every transcript for judge,
ground-truth and search references, which is where the "key 0, web 0" column comes
from. The published streams are exactly what those greps ran on, so the audit is
repeatable. Codex is the one CLI with no per-tool disable flag, and its transcript is
where the refused host-app and browser attempts appear.

**One addition you had asked for on the sister PR** and which applies here too:
`all_frames`, the all-frames-pasted / no-tools condition. The whole match goes into
context as 80 uniform stills and the shell is taken away — no seeking, cropping,
upscaling or scripting, only reading frames and writing an answer. It submitted 15
events and matched no rally anchor: **0.0**. It is in `calibration/ablations/` and
pinned in `test_judge.py` with the rest.

**Still open, and it is the Antigravity row.** A clean-room run is executing now under
the documented isolation — workspace outside the repo, `trustedWorkspaces` narrowed to
it so `--sandbox` refuses every other path, and the agent's persistent memory moved
aside beforehand — with the same integrity gate that voids the window on grounding, a
successful bypass, an answer copy, or a perfect claim on too few frames. I will post
its answer, score and raw trace whether it converges or not. If it comes back voided
by the gate rather than merely low-scoring, I will ask for the scoped waiver
explicitly rather than reporting a tainted number.

The per-field observability witness is the other one I still owe you; replying on that
thread separately.

techgenmini:
> Hi @luojiaxuan — thanks for publishing the streams. I rechecked ee66a0c: the GT rebuild and regression suite pass, the media-provenance bump is correct, and the pinned Codex/Opus/three original-ablation
  > archives verify. I found no successful web or answer-key lookup in the retained streams, so the remaining issue is environment parity, not an integrity accusation.
  >
  > The remaining work is bounded:
  >
  > - [ ] ee66a0c also adds the unrelated byu-wsu-2023-volleyball-ace-block-timeline/ task directory. Please remove it from this PR or split it out.
  > - [ ] The final Codex and Opus rows were not run with the shipped tool profile: the image supplies ffmpeg/curl/CA, while the traces use host Tesseract/ImageMagick/NumPy/Pillow; frame_dump also uses Pillow.
  >   Please rerun only those affected rows through the frozen task environment, respecting allow_internet=false, and archive the replacements under the same URL/SHA scheme. Alternatively, deliberately change
  >   the image tool profile and recalibrate that revised image. This is not an n=3 request; no need to repeat unaffected rows solely for this.
  >
  > - [ ] Add the native all_frames trace and its URL/SHA to the immutable archive, or withdraw that measurement until its 80-frame/no-shell setup is auditable.
  > - [ ] Complete the per-field observability ledger: the current sheets establish all score anchors, but not every blocker and blocked-hitter attribution. A direct crop/timestamp or a documented rotation-
  >   continuity chain is sufficient; otherwise relax unsupported fields.
  >
  > - [ ] Finish the Antigravity row with its answer, score, and raw trace, or request the scoped waiver described above.
  >
  > Once these are addressed, I do not see a reason to reopen the GT, scorer, model-lineup, or general ablation questions.

tasks/agentic_vbench_understanding/usc-wsu-2023-volleyball-block-timeline/calibration/observability/flips.json
luojiaxuan: Done, in `9e49836` — and the answer is not the one I expected, so here it is straight.

`calibration/observability/witness.md` now covers **every credited blocker and every blocked hitter on all 23 points**. Each row cites a frame strip in `observability/witness/`: four moments per event (t+0.8, +1.6, +2.4, +3.2 after the bug flips), cropped to the band the net sits in and kept at the source's horizontal resolution, so the numbers survive.

| | events |
|---|---|
| every credited blocker legible | **11 / 23** |
| some credited blockers legible | 4 / 23 |
| no credited blocker legible | 8 / 23 |
| blocked hitter legible | **1 / 23** |
| blocked hitter plausible but occluded | 2 / 23 |
| blocked hitter not legible | 20 / 23 |

**You were right to push on this.** The blockers hold up: on 11 events every credited name is readable, and where any number is legible it matches the answer key — no event contradicts the official record. **The blocked hitter does not.** On 20 of 23 there is no legible frame of that player in the window at all, and the reason is structural rather than incidental: the broadcast cuts to whoever is celebrating, which is the blocking side, so the stuffed hitter is out of frame or facing away. The eight events with nothing legible are the set-winning points, where the feed goes to the crowd or the changeover, plus the ones where the camera stays wide.

So on your standard — a timestamp/frame/crop witness per field — the blocker field passes and the hitter field does not. I am not going to argue it into passing. What I would say in its defence is that "not readable after the whistle" is not the same as "not recoverable": the hitter is the player whose attack the agent just watched, identifiable from the rally by position and rotation rather than by reading a number. And the scorer already treats the two fields asymmetrically — blockers exact with the hitter wrong earns partial credit, not zero.

Your call on which way to take it, and I will implement either:

1. **keep the field**, with `witness.md` shipping as the honest statement of what is directly readable; or
2. **relax it** — full credit on set, score_after and the blocker multiset, with the hitter dropped or moved to a bonus. That changes the scorer, so the calibration numbers would need regenerating; I would rerun rather than re-grade, which is a few hours.

One thing I fixed while building this, worth flagging because it is the file you commented on: `flips.json` was carrying the **coarse 20 s grid** rather than the bisected timestamps in `flips.md` — up to 17 s late on 16 of the 23 rows. It now holds the bisected values, each with a fresh read-back of the score bug at that exact second: 23/23 confirm.

tasks/agentic_vbench_understanding/usc-wsu-2023-volleyball-block-timeline/calibration/observability/README.md
luojiaxuan: Superseded by the per-field ledger in `calibration/observability/witness.md` (`9e49836`) — replying in full on the `flips.json` thread. Short version: every credited blocker is legible on 11 of 23 points, some on 4, none on 8; the blocked hitter is legible on 1. The set 1 / 14-10 case you cited is the one event where the hitter reads and the blocker does not, which turns out to be the exception rather than the pattern.

luojiaxuan:
First of the five, `49a9bb6`: **the BYU directory is out of this PR.** You are right and it was mine — the `ee66a0c` commit used a broad `git add` and swept in 29 files from the sister task's old path that were sitting untracked in the worktree. Removed, and the path is in `.gitignore` on this branch so it cannot come back. `gh pr diff 61 --name-only | grep byu` is now empty.

One thing you have not seen yet: the **per-field observability ledger landed 15 minutes after your comment**, in `9e49836`. `calibration/observability/witness.md` covers every credited blocker and the blocked hitter on all 23 points, each row citing a four-frame strip in `observability/witness/`. It does not come out the way I would have liked:

| | events |
|---|---|
| every credited blocker legible | **11 / 23** |
| some credited blockers legible | 4 / 23 |
| no credited blocker legible | 8 / 23 |
| blocked hitter legible | **1 / 23** |

Your "or a documented rotation-continuity chain" is the route I would take for the hitter, and I will build that rather than argue the crop standard into passing. Detail on that thread.

Working the rest in order: all_frames next, then the frozen-image reruns, then Antigravity.

luojiaxuan:
**all_frames is withdrawn and replaced.** You were right, and the trace shows it is worse than the answer alone: the model never opened a single frame. Its own closing note says so — no shell, no `Glob`/`LS`, `Read` returning `EISDIR` on the directory, ~16 filename conventions guessed and none resolved — and it labelled the submission "a structurally plausible fabrication". I had withheld the shell and the directory tools together, which left it no way to *discover* the filenames. That measured my harness, not the task. The same failure was in the USC leg: 17 of 22 tool results were errors.

The replacement fixes the three things you named:

- **the frames are reachable** — every filename is listed in the prompt, so no directory listing is needed;
- **the verifier and key are isolated** — the workspace is `~/avb-allframes/<task>/`, outside the repository, so no judge, answer key or sibling run is on a path the model could open;
- **the prompt is the frozen block-only instruction**, unmodified except for the ablation clause, and committed verbatim as `calibration/ablations/all_frames.instruction-as-run.md`.

Tools left available: `Read` and `Write` only. `Bash`, `Glob`, `Grep`, `Edit`, `MultiEdit`, `NotebookEdit`, `Task`, `WebFetch` and `WebSearch` are all withheld.

Result — and this time the model executed the condition:

| task | frames opened | events submitted | anchors matched | score |
|---|---|---|---|---|
| USC–WSU | 80 of 80 (80 `Read` calls, 1 error) | 10 | 2 | **0.0** |
| BYU–WSU | 76 of 77 | 12 | 1 | **0.0** |

It is worth reading what the USC model did with the frames, because it is the right negative result rather than a failure to try: it reconstructed the whole set-score progression from the on-screen graphic, got the final 3-2 and every set score right, and then explained why the condition cannot yield block points — at a 90 s stride a rally is essentially never captured mid-action, and a single blurry instant near the net cannot distinguish a stuff from a kill through the block, a deflection kept in play, or an attack error.

**Native traces published**, same scheme as the rest, at revision `c66a1ed95a6664c1f423cc0b8f1eee4d2f242e01`:

| task | file | sha256 |
|---|---|---|
| USC–WSU | [`all_frames.stream-json.gz`](https://huggingface.co/datasets/gavinlaw/agentic-vbench-calibration-trajectories/resolve/c66a1ed95a6664c1f423cc0b8f1eee4d2f242e01/usc-wsu-2023-volleyball-block-timeline/ablations/all_frames.stream-json.gz) (21.0 MB) | `e6132b8b3a73fca3dd9205a6108ec24e6d8f6facf456b96b6ce494338a30a3be` |
| BYU–WSU | [`all_frames.stream-json.gz`](https://huggingface.co/datasets/gavinlaw/agentic-vbench-calibration-trajectories/resolve/c66a1ed95a6664c1f423cc0b8f1eee4d2f242e01/byu-wsu-2023-volleyball-block-timeline/ablations/all_frames.stream-json.gz) (17.0 MB) | `76afe71d06f59262629b8c00a71e069551ae3abd9dbf534a69d5261ec171d0c3` |

The exact prompt for each is in the same dataset directory. Both rewards are pinned in `test_judge.py` with the rest.

Next up is the environment-parity rerun; the revised image is built and the wrapper is written.

luojiaxuan:
Environment parity is done for this task, in `7ee65f3`, and it moved a number — upward.

**I took your second option: change the image tool profile deliberately, then recalibrate against it.** The traces showed the agents reaching for Tesseract, ImageMagick, NumPy and Pillow, none of which the image shipped. Rather than measure agents crippled relative to what a video task obviously needs, the image now ships that working set, with the reasoning in the Dockerfile. If you would rather the profile stay minimal, say so and I will rerun against the lean image instead — that is a rerun, not a rewrite.

**The confinement is structural, not instructed.** The workspace the agent sees holds no video at all; the broadcast exists only inside the container, and the only route to it is a three-line wrapper:

```sh
exec docker exec -w /workspace/work avb-parity-<task> sh -lc "$*"
```

The container runs `--network none`, so no task action reaches the network. `/workspace/work` is bind-mounted so frames extracted inside are visible to the agent outside. Recorded in `calibration/parity/environment.txt`:

```
image digest: sha256:58ed2ce8cbd8a220b9d2c6249bac619955d95de611a4097e35a12834332e1aa2
container network: none
media inside the image: 13ccbabb…d08ba9e   (matches the pin)
ffmpeg, ffprobe, tesseract, convert, python3 + numpy 2.1.3, Pillow 11.0.0
```

**Results.**

| agent | in the task image | on the host |
|---|---|---|
| Codex CLI, gpt-5.6-sol xhigh | **0.0789** | 0.0185 |
| Claude Code, Opus 5 xhigh | **0.0** | 0.0 |

Codex is higher confined than on the host, and I want to be straight about it rather than bury it: 0.0789 against a ~0.10 bar is thinner than the earlier number implied. It comes from precision, not from getting blocks right — it submitted 15 events instead of 31, matched 5 rally anchors and earned 3 partials, and still got **no block point fully correct**.

**Opus's run is worth your time.** It submitted an empty list and wrote up why, committed verbatim as `calibration/parity/opus.agent-notes.md`. In 242 tool-call turns it rebuilt **196 of the match's 200 points** from the score bug — including both overturned calls — and anchored 168 rallies to a camera cut within 0.5–3.5 s of the ball landing. Then it stopped:

> the ball is ~8–14 px and often motion-blurred […] at 0.1 s sampling the ball is readable, but each such sheet covers only ~1.2 s, so it needs an anchor tighter than the ±3.5 s I can derive […] The broadcast shows no replays.

It refused to ship low-confidence guesses because the instruction says to list only genuine block points. That is the intended failure mode, from an agent that had OCR and NumPy available.

One thing in it bears directly on your observability thread: it reports that **blocker and hitter jersey numbers are readable from the live wide shot at 4× upscale** (it confirmed USC libero #16). So what defeats it is classifying the rally ending, not identifying players — which is the rotation-continuity route you offered, arrived at independently.

**Native traces published** at revision `c88071f7988507cc81fdb83e7c0b81ce25b8e1cc`, with the exact prompt and `environment.txt` beside each. Every SHA256 in `scores.md` is checked against what the dataset actually serves — 10 of 10 match.

Next: the same treatment for the sister BYU task, then the setter ledger. Antigravity is still waiting on quota; its first window came back CLEAN but exhausted before committing an event, and a resume attempt hit 429 immediately.

techgenmini:
Hi @luojiaxuan — I checked the published parity trace here as well, and the same confinement issue applies.

  The original MP4 is accessed through `./run`, but substantive work on the bind-mounted derived frames still executes through native host tools. For example, the Codex stream runs `find_rally_ends.py` with
  host NumPy/Pillow and performs additional image processing from the host shell; only part of the extraction/OCR pipeline goes through the container. The resulting `0.0789` / `0.0` rows are useful exploratory
  evidence, and I found no source lookup, but they do not yet establish shipped-image/network parity.

  The richer Docker tool profile itself is fine; no lean-image rerun is requested. Please first confine all Python/OCR/image-processing actions to the no-network container, then run one corrected Codex trial.
  Do not spend on another Opus or Antigravity attempt yet; after the corrected gate-setting row, we can decide whether the audited Opus zero and quota-limited Antigravity attempt warrant scoped acceptance
  rather than another expensive run.

  The attribution/rotation-continuity evidence remains the separate task-design item already under discussion.

luojiaxuan:
Corrected Codex trial is done, `16f67a2`. **0.0**, down from the 0.0789 the flawed harness produced. Reporting it as it landed.

**The agent now runs inside the task image**, not behind a wrapper. The CLI itself executes in a container built from `environment/Dockerfile`, so every command it issues necessarily uses that image's tools; there is no host tool surface for it to reach. The prompt is the shipped instruction verbatim — `parity/instruction-as-run-v2.md` hashes to the same SHA256 as `steps/solve/instruction.md` — because inside the environment the video already sits where `workdir/setup.sh` puts it.

**Egress is default-deny.** `OUTPUT` policy DROP; the one hole is an allowlisting CONNECT proxy that takes the CLI's backend and refuses every other host by name. Verified against the running container:

| target | result |
|---|---|
| `stats.ncaa.org` (the answer source) | refused by the proxy, logged |
| raw `1.1.1.1:443`, bypassing the proxy | dropped by iptables, no connection |
| the model backend | allowed |

```
harness digest: sha256:8f10925ce436f30342104a876117dda2dfbf5e92b5c9443f54688aa9e010f97e
task image under it: avb-usc-parity:latest
agent: codex-cli 0.151.0, inside the container
media inside: 13ccbabb…d08ba9e  (matches the pin)
tools the agent has: ffmpeg, ffprobe, tesseract, convert, python3 + numpy 2.1.3, Pillow 11.0.0
                     (cv2, scipy absent)
```

**Result: 0.0.** 16 events over 153 command executions and three hours ten minutes; 2 of 23 rally anchors; no block point fully *or* partially correct.

| | old harness | corrected |
|---|---|---|
| reward | 0.0789 | **0.0** |
| events | 15 | 16 |
| rally anchors | 5 / 23 | 2 / 23 |
| fully correct | 0 | 0 |
| partials | 3 | 0 |

The drop is in *anchoring* rallies, not in attribution — attribution was already at zero. Host tooling made a batch OCR pass over the score bug cheap; inside the image the same work costs more, so far fewer rallies get anchored and the three partials go with them.

**The egress log is published with the trace**: 994 connections to the model backend, **zero to anything else**, and zero refusals — the agent never asked. That is a proxy's record rather than a grep over a transcript, and it is hashed in `scores.md` like the rollout. Both hashes were re-checked against what HF actually serves.

The run ended on `turn.completed` with its own validation step, not cut off. It hit five WebSocket reconnects mid-run and fell back to HTTPS transport; that is in the published stream.

`test_judge.py` now pins this reward, so the number is a test failure if it ever drifts from the committed scorer.

**Opus and Antigravity are untouched**, as you asked. The Opus parity row is marked *provisional* in `scores.md` and `SPEC.md` — measured under the flawed harness, not rerun. Ready for whatever call you want to make on those two given a corrected gate-setting row at 0.0.

techgenmini:
Thanks again for the unusually thorough work on this task. I have merged the PR at `16f67a2`. This is post-merge follow-up tracking, not a request to revise or reopen the merged PR.

When you have time, could you please open one follow-up issue and link it here? I would like it to track this bounded set:

- **Attribution observability / scored-field contract.** The current witness directly resolves every credited blocker on 11/23 events, some blockers on 4/23, and no blocker on 8/23; the blocked hitter is directly legible on 1/23. Please either add a row-level contact/rotation-to-identity chain for the fields that remain scored, or mask/drop unsupported fields or events, then align `witness.md`, `SPEC.md`, the instruction, and the judge. A general statement that identity can be inferred from rotation is not a substitute for the per-row chain.
- **Final calibration disposition after that contract is frozen.** The corrected in-image Codex row at 0.0 is accepted. The Opus row remains provisional under the earlier host-processing harness, and Antigravity remains not run. Please use the issue to record corrected final rows or request explicit scoped waivers; rerun Codex and affected ablations only if the prompt/scorer contract changes.
- **Small evidence cleanup.** Bind the retained final run to the immutable task/base-image digest rather than only the `latest` tag, and correct the stale `all_frames` count in `SPEC.md` and `calibration/ablations/README.md` from 15 submitted events to the actual 10.

The GT derivation, scorer regression coverage, immutable media, raw-stream archive, replacement all-frames trace, and corrected Codex Docker/network confinement are closed and do not need to be reopened in that issue.

No urgency—please link the follow-up issue here when it exists so we have a durable place for the remaining decisions.

luojiaxuan:
Follow-up issue is #119, tracking the three items: attribution contract, calibration disposition after it freezes, and the evidence cleanup.

On the cleanup, both points check out: `environment-v2.txt` pins the harness image by digest but the task image only by `:latest`, and the `all_frames` count is 10 in both `all_frames.solution.json` and `all_frames.reward.json` against 15 in the two docs. The other three ablation counts are correct.

PR #77, opened 2026-07-26
Super Bowl LI referee-announced player-foul timeline task

### LINKED ISSUE #60: [Task Proposal] Super Bowl LI referee-announced player-foul timeline
Task Proposal — Super Bowl LI referee-announced player-foul timeline
Family: agentic_vbench_understanding
Proposed task id: ne-atl-2017-super-bowl-li-penalty-timeline

One-line
Given the full 143-minute broadcast of Super Bowl LI, reconstruct every referee-announced player foul: quarter, game clock, infraction type, the jersey number of the penalised player, and their team.

Video
Full broadcast of Super Bowl LI (New England Patriots vs Atlanta Falcons, 5 February 2017), archived on archive.org:
https://archive.org/download/youtube-noLK78Hgq0A/noLK78Hgq0A.mp4
1920x1080, 2:23:00, 4.0 GB, direct file URL, verified seekable by HTTP range request.
The task pins this upstream file by URL and SHA-256 and bakes it at build time. No re-hosting is performed. This is the same posture as the gsw-cle-2018-finals-g4-three-point-timeline worked example, which bakes an archived broadcast from archive.org.

Question and output
The agent receives:

/workspace/materials/game.mp4

The agent writes /workspace/output/solution.json:

{
  "penalties": [
    {
      "quarter": 3,
      "clock": "1:25",
      "type": "offensive holding",
      "player_number": 70,
      "team": "ATL"
    }
  ]
}

quarter is 1 to 4, or 5 for overtime. clock is defined precisely below. type is drawn from a closed vocabulary supplied in full in the instruction. player_number is the jersey number of the penalised player. team is NE or ATL. One entry per scored foul, in any order.

Scope rule (mechanical): a scored event is a foul for which the referee announces a jersey number. Team fouls that carry no announced number — delay of game, illegal formation, and similar — are excluded. The reason is given under Anti-shortcut: they are recoverable from video alone and would leak true positives.

Clock rule (operational): the clock value is the game clock shown in the score bug on the last frame it is displayed before the referee begins the announcement. The broadcast hides the score bug during live action and restores it between plays, so this rule names one unambiguous frame rather than leaving the agent to choose among several plausible displayed times.

Ground truth (machine-truth)
The primary source is the NFL's official Game Book for this game, published through the NFL Game Statistics and Information System (nflgsis.com) and linked from the NFL.com Game Center. Game Books are the league's official record and carry a full play-by-play with a penalty summary: penalised player, infraction, yardage, enforcement, quarter, and game clock.

Ground-truth construction is deterministic:

Parse the penalty entries from the official Game Book play-by-play.
Map each penalised player's name to their jersey number using the roster printed in the same Game Book.
Apply the scope rule, retaining only fouls with an announced jersey number.
Normalise infraction names to the closed vocabulary supplied to the agent.

No human labeling supplies identity, timing, or infraction type. The record is independently cross-checkable against the per-game penalty data published by nflpenalties.com, which lists 13 penalties assessed for 88 yards and 16 flags overall for this game.

Why cross-modal — as measured, not asserted
We sampled the actual broadcast by remote-seeking the archived file and inspecting frames. The findings changed the design and are reported here rather than assumed.

What the video supplies:
The score bug shows quarter, game clock, score, and down and distance. It is displayed continuously between plays and hidden during live action. Sampled at 3RD 2:05, 2:01, 1:35, 1:25, 1:22, 1:09, :59, :55 — the clock is readily readable.
When a flag is thrown, a generic yellow FLAG banner appears. It carries no type and no number.
After the referee's announcement, a banner appears with the offending team's logo and the infraction type — for example, a Falcons logo above DELAY OF GAME. The infraction type is therefore burned into the picture.

What the video does not supply:
No jersey number appears in any graphic. Across twenty sampled crops of the graphic zone spanning a flag, an announcement, and the surrounding play, no number was displayed at any point.

What the audio supplies:
The referee's microphone states the infraction and the jersey number: "Holding, offense, number 76."

What the audio does not supply:
The referee never states the game clock.

Consequently the infraction type and team leak to OCR and carry no difficulty. The task does not rest on them. The compound match rests on two fields that live in different channels and nowhere else: the jersey number, available only in the referee's audio, and the game clock, available only in the on-screen graphic. Neither channel alone can produce a single true positive.

Why hard and long-horizon
Full-video coverage: fouls are scattered across a 143-minute broadcast with no graphic listing them; a complete answer requires traversing the whole game.
No text channel contains the answer: the broadcast carries no penalty list, and transcribing commentary does not yield the compound record — commentators discuss penalties inconsistently and rarely state the jersey number and the game clock together.
Compound exact match: a true positive requires the infraction type, jersey number, team, quarter, and clock together. Detecting that a flag was thrown, or OCR-ing the type banner, is insufficient.
Semantic, not signal: the referee's announcement must be understood and bound to a graphic that is not on screen at the same moment. There is no signal-processing shortcut that produces the record.

Scorer
Pure-Python event-level F1 with greedy one-to-one matching, deterministic, stdlib only.

A predicted foul is a true positive only when:

type matches after normalisation;
player_number matches exactly;
team matches;
quarter matches exactly;
clock is within 5 seconds of the official game clock;
the ground-truth event has not already been matched.

Duplicate predictions are false positives; unmatched ground-truth fouls are false negatives. The reward is the harmonic mean of precision and recall.

Oracle = 1.0. Empty/null = 0.0. A guessed list scores approximately 0, because the compound match requires the jersey number and the clock simultaneously.

Anti-shortcut
Video-only: OCR recovers type, team, and clock, but never the jersey number, which appears in no graphic. Guessing a number from roughly one hundred dressed players makes a compound true positive vanishingly unlikely; expected score near 0.
Audio-only: the referee states the infraction and the number but never the game clock, so no foul can be placed; expected score near 0.
Team fouls excluded: delay of game and similar fouls carry no announced number, so all their fields are recoverable from the graphic alone. Left in scope, a video-only run could match them and clear the 0.15 ceiling on those events. The scope rule removes them. This is a direct consequence of the frame evidence above.
Single frame: cannot reconstruct a multi-event ledger across 143 minutes, and carries no number.
Frame dump without tools: frames carry no referee audio, so the jersey number cannot be recovered by one inference pass over pasted images.
No media: the compound ledger of fouls with jersey numbers and clock times is not recallable and the schema is not guessable. This is the ablation to scrutinise, because Super Bowl LI is a famous game. The worked NBA example is also a famous game, and its README records an earlier box-score formulation scoring 0.64 by recall before the timeline redesign killed that shortcut. The outcome of this game is famous; its penalty ledger is not. To be measured.

Feasibility status
No pilot has been built. The following separates what has been verified from what has not.

Verified by direct inspection of the archived file:

the source exists, is 1920x1080, 2:23:00, and is remotely seekable for frame extraction without a full download;
resolution and duration satisfy the 720p floor and the 10-300 minute window;
the score bug is intermittent — hidden during live action, restored between plays — which is why the clock rule names a specific frame;
a flag produces a generic FLAG banner carrying no type and no number;
an announcement produces a team-logo banner carrying the infraction type, so type and team leak to OCR;
no jersey number appears in any sampled graphic.

Verified from official and cross-check sources:

NFL Game Books are live and served as real PDFs from nflgsis.com;
this game had 16 flags and 13 accepted penalties for 88 yards.

Not yet verified:

whether the referee's announcements are consistently audible and intelligible over crowd noise — this is the design's single point of failure and the first thing the pilot will test;
the exact Game Book URL for this game; the nflgsis.com path scheme is confirmed but the game identifier is not;
how many of the 16 flags survive the scope rule; the expectation is roughly 12-13, which is thin;
oracle, empty, and ablation figures above are design expectations, not measurements;
strong-agent difficulty below 0.10 and rollout length above 50 tool-call turns are unmeasured.

Open questions
Event count. After excluding team fouls, roughly 12-13 events remain, against 22 in the worked NBA example. Thin counts make F1 noisy. Levers: widen to every referee announcement including replay reviews and measurements, which the Game Book also records; or select a higher-penalty game — Super Bowl LV (archive.org youtube-arcwCWZBY8Q, 1920x1080, 2:04:00) was notably flag-heavy. Which do you prefer?
Referee-audio dependence. The entire cross-modal claim now rests on the jersey number being audible in the archived broadcast, since type and team are conceded to OCR. If the number is unintelligible under crowd noise on some fouls, those events are recoverable from neither channel and the oracle cannot reach 1.0. Should such fouls be excluded by a stated rule, or does that make the scope rule too bespoke?
Rights. The archive.org item carries a CC BY 4.0 tag, but this is very likely invalid — the uploader is not the rights holder and cannot relicense an NFL broadcast. We treat the footage as unlicensed and pin the pre-existing archive.org file rather than re-hosting, matching the gsw-cle worked example. Is that the posture you intend for archived broadcast media?
Fame. Archived full NFL broadcasts at 720p or above appear to be exclusively Super Bowls; obscure games survive only as sub-720p transfers and fail the resolution gate. If less-famous footage is required, this constraint may be binding.

Plan
Verify referee-audio intelligibility across a sample of fouls — the design's single point of failure — before anything else.
Locate and pin the official Game Book; parse the penalty summary and roster; apply the scope rule; emit the oracle.
Build a one-quarter pilot; validate oracle 1.0 and empty 0.0.
Run the video-only ablation early, since type and team are known to leak to OCR and the design depends entirely on the jersey number not leaking.
Run the audio-only and no-media ablations.
Package the task in the Harbor agentic_vbench_understanding layout; run check_task.py and all required anti-shortcut ablations.
Calibrate with Codex, Antigravity, and Claude Code, retaining raw trajectories.
Iterate on scope until a strong-agent run scores below 0.10 and exceeds 50 meaningful tool-call turns.
Submit the completed task, filled SPEC.md, scorer, oracle, calibration table, and rollouts as a PR.

YashJayswal24:
## Task: Super Bowl LI referee-announced player-foul timeline

Proposed and discussed in #60. Adds `tasks/agentic_vbench_understanding/ne-atl-2017-super-bowl-li-penalty-timeline/`.

**Opened for design/code review — NOT yet calibration-complete, so it should not be merged until the calibration lands.** Shared early per the proposal-first process, so the design, scorer, and ground truth can be reviewed before the (expensive) calibration stage.

### What the task is
Reconstruct every referee-announced player foul from the full Super Bowl LI broadcast: quarter, game clock, infraction type, penalised player's jersey number, team. Cross-modal by construction — the jersey number is stated only by the referee's mic (audio), the game clock only by the on-screen score bug (video); neither channel alone yields a true positive. Ground truth = official NFL Game Book (machine-truth). Media pinned from archive.org (pin, not re-host), matching the `gsw-cle` worked example.

### Verified so far (locally)
- **Media:** `youtube-noLK78Hgq0A.mp4`, 1920×1080, 2h23m; **SHA256 pinned** in the Dockerfile (`ba2281d6…4ba998`, computed by streaming the file).
- **Scorer (`judge.py`):** deterministic compound-match F1. Local pilot: **oracle → 1.0, empty → 0.0, plausible guess → 0.0**, partial (clock off 3 s) → 0.57. Clock tolerance and wrong-number rejection behave as designed.
- **`no_media` gate (maintainer's step 1):** indicative proxy run — a no-media, no-tools model reconstructed **0/13** penalties (F1 0.0). De-risks the fame concern; the official GPT 5.6 Sol run is still required.
- **Audio intelligibility (design's single point of failure):** the referee's announced jersey number transcribed cleanly on a small CPU ASR model for **4/4** sampled player fouls — No. 70 Matthews (Q3), No. 23 Alford (Q2), No. 34 Poole (Q2), No. 59 Campbell (OT PI).

### Now complete (update)
- **Full ground truth — DONE.** Replaced the 4-foul pilot with **13 referee-announced player fouls** parsed from the official NFL Game Book play-by-play; jersey numbers taken from the Game Book lineups (the 4 audio-verified numbers all match). `judge.py` GROUND_TRUTH and the oracle `solve.sh` are in lockstep — locally **oracle 13/13 → 1.0, empty → 0.0, partial → 0.83**. Scope rule + the 3 no-number exclusions (delay of game, illegal formation, illegal touching) documented in `PROVENANCE.md`, per @christine1729's guidance in #60.

- **`check_task.py` — PASSES (partial).** Ran the repo's own `scripts/understanding/check_task.py`: **PASS** on task structure, input video (**143.1 min, 1080p** — resolution now confirmed by its ffprobe, not assumed), oracle == 1.0, baseline == 0.0. The agent-reward / agent-turns / ablation inputs are reported **SKIP** (they need the strong-agent stack).

### Still pending (needs the agent/Harbor stack — same dependency as calibration)
- **Calibration.** GPT 5.6 Sol / Codex / Antigravity rollouts + performance table + raw trajectories under `calibration/`. `scores.md` marks these `_to run_`.
- **Ablations.** `single_frame`, `video_only`, `audio_only`, `frame_dump_no_tools` — each is a strong-agent run on degraded media, so it needs the same stack as calibration.

### Open design questions on the ground truth (see `PROVENANCE.md`)
- **Declined fouls** (3 of the 13) are included because the referee announces the number; accepted-only would give 10. Your call.
- **Event count** is 13 vs 22 in the NBA example; widening to replay reviews / measurements would need a schema addition, so it is left open rather than silently included.

### Review asks
1. Is the scope rule (referee-announced *player* fouls; team fouls excluded because their fields are OCR-recoverable) the right boundary?
2. With type + team conceded to OCR, is the compound-match-on-number+clock cross-modal argument sufficient, or do you want a stricter formulation?
3. Anything you'd want changed in the task layout / scorer before the calibration spend?

Closes #60 once calibration lands (not with this draft).

### DIFF
--- a/tasks/agentic_vbench_understanding/ne-atl-2017-super-bowl-li-penalty-timeline/PROVENANCE.md
+++ b/tasks/agentic_vbench_understanding/ne-atl-2017-super-bowl-li-penalty-timeline/PROVENANCE.md
@@ -85,20 +85,21 @@ carry a player jersey number, by rule/convention:
 | 3 | 2:06 | NE | illegal touching (kick) | S. Gostkowski | 3 | kicking-team touching — announced without a player number |
 | 3 | 0:04 | ATL | delay of game | M. Bosher | 5 | team foul — no player number |
 
-## Open design questions (for the PR / issue #60)
-- **Declined fouls (rows 2, 3, 8):** included because the scope rule says the referee
-  announces the number regardless of the penalty decision. That assumption is verified
-  by transcription for 0 of these 3 rows so far (see the audio-verified column above) —
-  still open until confirmed the same way the 4 accepted rows were.
-- **Event count:** 13 natural events (16 total referee-mic penalty announcements, minus
-  3 with no jersey number). Reaching the ≥20 figure from review ask #4 would require
-  adding replay reviews and/or measurements, but this game has at most 1 review and no
-  measurements (see enumeration table above), and neither carries a jersey number or an
-  audio-only component — padding with them would dilute the audio/video seam rather than
-  extend it. Recommendation: keep the ~13-event scope and address the F1 fragility
-  (review ask #4's "single lucky guess clears 0.10" concern) as a scoring-shape question
-  instead of an event-count question.
-- **Clock precision:** superseded by the observable clock rule above (per review ask #5)
-  — clocks are the official Game Book values, scorer tolerance ±5 s, 3 of 13 rows
-  spot-checked against the score bug by frame extraction; full per-row audit against the
-  rule is open.
+## Scope and scoring ruling (settled by techgenmini on issue #60)
+- **Scope:** keep the natural ~13 referee-announced player-foul events. Do not pad with
+  replay reviews or measurements — they carry no jersey number and no audio-only
+  component in this game, so adding them would dilute the audio/video seam rather than
+  extend it.
+- **Declined fouls (rows 2, 3, 8):** stay in scope only if each one's number is
+  individually confirmed audible. Verified by transcription for 0 of these 3 rows so
+  far (see the audio-verified column above) — pending, per-row, not a blanket rule
+  anymore.
+- **Scoring:** F1 replaced with F2 (β=2, recall-weighted) in `judge.py` — a single lucky
+  exact row no longer clears the 0.10 anti-shortcut gate (F1 gave 0.1429; F2 gives
+  0.0943). Regression-checked against null/one-hit/two-hit/oracle in
+  `steps/solve/tests/test_regressions.py`.
+- **Clock rule:** the observable rule above (last score-bug frame before the
+  announcement) is now the binding definition. Scorer tolerance stays ±5 s. 3 of 13 rows
+  were spot-checked against the score bug by frame extraction during initial design;
+  the full per-row observability audit against this rule — for all 13 rows, alongside
+  each row's audio timestamp/transcript and the Game Book clock — is still pending.
--- a/tasks/agentic_vbench_understanding/ne-atl-2017-super-bowl-li-penalty-timeline/PROVENANCE.md
+++ b/tasks/agentic_vbench_understanding/ne-atl-2017-super-bowl-li-penalty-timeline/PROVENANCE.md
@@ -1,0 +1,54 @@
+# Ground-truth provenance — Super Bowl LI penalty timeline
+
+## Sources
+- **Penalty list + game clocks:** official NFL Game Book play-by-play for Super Bowl LI
+  (NE 34–28 ATL, 2017-02-05). Cross-checked against nflpenalties.com (13 accepted
+  penalties, 88 yards; 16 flags total).
+- **Jersey numbers:** the same Game Book's Lineups and Substitutions section (authoritative
+  official roster). No jersey number is taken from memory or inference.
+- **Independent audio confirmation:** four fouls had their referee-announced number verified
+  by transcribing the broadcast audio — #23 R. Alford, #34 B. Poole, #70 J. Matthews,
+  #59 D. Campbell. All four matched the official lineup.
+
+## Scope rule (mechanical)
+A **scored event** is any penalty in the Game Book play-by-play whose infraction is an
+**individual player foul that the referee announces with a jersey number** — whether the
+penalty was accepted or declined (the referee announces the number either way). The jersey
+number is the player's number in the official Game Book lineup.
+
+## The 13 scored fouls
+| # | Q | clock | team | type | player | № | accepted? |
+|---|---|-------|------|------|--------|---|-----------|
+| 1 | 1 | 13:47 | ATL | offensive holding | P. Worrilow | 55 | yes |
+| 2 | 2 | 14:19 | NE | offensive holding | M. Bennett | 88 | declined |
+| 3 | 2 | 8:55 | NE | defensive pass interference | P. Chung | 23 | declined |
+| 4 | 2 | 8:02 | ATL | defensive holding | R. Alford | 23 | yes (audio ✓) |
+| 5 | 2 | 6:10 | ATL | defensive holding | B. Poole | 34 | yes |
+| 6 | 2 | 5:16 | ATL | defensive holding | B. Poole | 34 | yes (audio ✓) |
+| 7 | 2 | 0:18 | NE | offensive holding | M. Bennett | 88 | yes |
+| 8 | 3 | 13:02 | NE | offensive pass interference | C. Hogan | 15 | declined |
+| 9 | 3 | 8:43 | NE | defensive pass interference | M. Butler | 21 | yes |
+| 10 | 3 | 1:30 | ATL | offensive holding | J. Matthews | 70 | yes (audio ✓) |
+| 11 | 4 | 3:50 | ATL | offensive holding | J. Matthews | 70 | yes |
+| 12 | 4 | 0:57 | ATL | defensive offside | D. Freeney | 93 | yes |
+| 13 | 5 (OT) | 11:18 | ATL | defensive pass interference | D. Campbell | 59 | yes (audio ✓) |
+
+## Exclusions (documented, per maintainer guidance in issue #60)
+These Game Book penalties are **not scored** because the referee's announcement does not
+carry a player jersey number, by rule/convention:
+
+| Q | clock | team | type | player | № | reason excluded |
+|---|-------|------|------|--------|---|-----------------|
+| 2 | 8:48 | NE | illegal formation | S. McClellin | 58 | formation foul — announced without a player number |
+| 3 | 2:06 | NE | illegal touching (kick) | S. Gostkowski | 3 | kicking-team touching — announced without a player number |
+| 3 | 0:04 | ATL | delay of game | M. Bosher | 5 | team foul — no player number |
+
+## Open design questions (for the PR / issue #60)
+- **Declined fouls (rows 2, 3, 8):** included here because the referee announces the number.
+  If the maintainer prefers accepted-only, dropping them yields 10 events.
+- **Event count:** 13 (vs 22 in the NBA worked example). The maintainer suggested widening to
+  *every* referee announcement including replay reviews and measurements; those carry no jersey
+  number and would need a schema addition, so they are left as a further extension pending
+  agreement rather than silently included.
+- **Clock precision:** clocks are the official Game Book values; the scorer allows ±5 s. Three
+  were additionally confirmed against the on-screen score bug by frame extraction.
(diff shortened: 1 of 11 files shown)


techgenmini:
> Thanks for opening this pre-calibration. Please respond to the scope/clock audit request on #60 before spending on calibration. I’ve also left inline comments for the leaked examples and clock contract.

YashJayswal24:
Pushed fixes for the two concrete bugs (leaked GT examples in `instruction.md`, `allow_internet`) plus the restored clock rule — commit 1bb2aaf. Full scope/clock/audibility audit posted on #60: https://github.com/PhiloLabs/agentic-vbench/issues/60#issuecomment-5451143791

techgenmini:
> Hi @YashJayswal24 — thanks, this audit is enough to settle the task shape before calibration.
  >
  > - Keep the natural 13 referee-announced player-foul scope. Please do not pad it with reviews or measurements: they lack the audio/video jersey-number seam that makes this task valuable. A declined foul
  >   stays only if its number is individually audible.
  >
  > - Raw F1 cannot remain: one exact row currently scores 0.1429. Please use recall-weighted F2 instead of an arbitrary two-hit cliff. It keeps continuous partial credit, gives one exact row 0.0943, two exact
  >   rows 0.1852, and oracle 1.0. Add null / one-hit / two-hit / oracle regressions.
  >
  > - Complete a compact observability audit for all 13 rows before runs: the spoken-number timestamp and transcript, plus the last pre-announcement score-bug frame, displayed clock, Game Book clock, and delta.
  >   Drop any row that fails. Pin the exact official Game Book PDF URL, SHA256, and row references. This is best described as official-record-derived GT plus a broadcast audio/video audit.
  >
  > - Two small prompt fixes: the prose example “Holding, number 70, offense” is still a real GT pattern; make it generic. Also replace “clock at the moment the foul occurred” with “clock defined below,” since
  >   the binding rule is the pre-announcement score-bug frame.
  >
  > - The spoken number is load-bearing, but the image currently ships only ffmpeg while the evidence used local Whisper. Either bake/pin the intended ASR path, or demonstrate the permitted exact-image setup in
  >   the final traces. Keep allow_internet=true per the existing ruling; raw traces must distinguish dependency/model downloads from prohibited source lookups.
  >
  > Once those land, run final-prompt no-media and modality ablations plus the required Codex CLI, Claude Code, and Antigravity traces. No n=3 campaign is needed unless a final result is near the gate.

YashJayswal24:
Pushed the rest of the punch list from the latest review — F1→F2 in `judge.py` (with checked regressions), the remaining prose leak ("number 70" example), and `faster-whisper` baked into the Dockerfile (build + offline transcription verified locally with `--network none`). Commit d7e5d1d. Details on #60: https://github.com/PhiloLabs/agentic-vbench/issues/60#issuecomment-5476519947

YashJayswal24:
@techgenmini can you check it out

techgenmini:
> Thanks — I checked d7e5d1d. The F2 implementation and expected regression values are coherent, the remaining prompt leak and clock wording are fixed, and the Dockerfile now includes the intended faster-
  > whisper setup. This was a static review; I did not independently rerun the Docker build or offline transcription.
  >
  > I also see from your #60 update that the exact Game Book pin/references, full 13-row observability audit—including the three declined fouls—and final-prompt calibration are still in progress. Because that
  > audit may change the retained event set and scorer, I’ll treat this as a good progress checkpoint rather than a merge-ready re-review and wait for the completed batch.
  >
  > One small reproducibility item to fold in before final runs: small.en is currently cached without an exact Hugging Face revision. Please pin its snapshot, or freeze and record the exact final image digest
  > used for calibration.

PR #100, opened 2026-08-01
Add Medvedev–De Minaur break-point ledger task

inFaaa:
## Summary

Adds `medvedev-de-minaur-2023-us-open-break-point-ledger`, a video-understanding task that reconstructs every break-point opportunity in Daniil Medvedev vs. Alex de Minaur at the 2023 US Open.

The task requires a chronological JSON ledger of all 16 opportunities, including player-specific score, server, first-serve status, outcome, serve direction, rally count, and terminal-stroke attributes.

## Media reproducibility

The official US Open YouTube page still plays 720p, but its current high-resolution downloads require a short-lived per-video PO token. The task therefore bakes a verified 1280x720 H.264/AAC reconstruction from a pinned public release asset with `curl` and SHA-256 verification:

`d78c9246d5dd36b812c71b5f39bfa43ab86d1a7adf711fb7bbc6ff1d66d618b2`

The Dockerfile follows the direct-download plus hash-check pattern used by `gsw-cle-2018-finals-g4-three-point-timeline`; it has no `yt-dlp` fallback.

## Ground truth and scoring

- 16 opportunities: Medvedev 5/10 and de Minaur 2/6 in the official match feed.
- Scores and outcomes cross-checked against Match Charting Project point logs.
- Deterministic strict exact-event F1 scorer; malformed or duplicate entries cannot increase reward.
- Oracle reward: `1.0`; empty ledger reward: `0.0`.

## Calibration and remaining review work

Raw Codex, Claude, Gemini, and access-blocked Fable trajectories are included. The spec transparently records that an independent full-video annotation and the required anti-shortcut ablations are still outstanding, so this is opened as a draft for review rather than presented as final-ready.

## Validation

- `scripts/understanding/check_task.py --task-dir ...` passed.
- Oracle scorer check: `1.0`.
- Empty-answer scorer check: `0.0`.
- Re-downloaded the public release URL with `curl`; SHA-256 and 1280x720 probe both matched.
- Docker daemon was unavailable locally, so a full `docker build` has not been run in this checkout.


christine1729:
Hi @inFaaa, thanks for the PR, a break-point ledger is a nice long-horizon cut of a match!

before merge:
- [ ] Measure the ablations, especially no_media.

PR #101, opened 2026-08-02
Add dual-runway operations ledger task

XuanmiaoG:
## Task: dual-runway operations ledger from night PTZ video + tower radio

Adds `tasks/agentic_vbench_understanding/dualrunway-ops-ledger/`.

### What the task is

Reconstruct the ledger of runway operations at a busy dual-runway complex over a
2.5-hour night recording: for each landing and takeoff, the video-timeline moment, the
operation type, the aircraft family, and the ICAO callsign.

**Cross-modal by construction.** The callsign is spoken only on the tower frequency; the
moment, the landing/takeoff distinction and the aircraft family exist only in the
picture. Neither channel alone yields a true positive. The two files also carry **no
shared clock** — `runway.mp4` starts 105 s into the capture, `tower.mp3` starts at 0 —
so the correspondence has to be recovered from content.

The camera is a hand-operated PTZ: it follows individual aircraft, wanders off to the
city between them, and does not catch everything that uses the runway. Its on-screen
tracking panel, which names the aircraft and states its type, is masked to black for the
entire video — otherwise both scored fields would be free.

### Why a constant time-shift does not solve it

Measured on this recording: the gap between a clearance being spoken and the operation
happening has **median 197 s but σ = 97 s**, spread 2–452 s, while ground-truth
operations are only **119 s apart** at the median. The jitter is the same order as the
spacing, so lining up the clearance sequence and applying one offset lands on the
neighbouring aircraft. Every row has to be confirmed against the picture individually.

Quantified: a naive pipeline given the **video channel as an oracle** (ground truth's
own times, operations and types), handed the airline prefix for free, and taking the
callsign from the nearest clearance in the preceding 400 s scores **0.0816 — 4 of 49**.
For contrast the correct flight number appears in that same window for **39 of 49**. The
information is in the audio; naive pairing extracts almost none of it.

### Ground truth

49 operations. Built from a live OpenSky state-vector capture run throughout the
recording (1864 successful polls, ~5 s cadence), joined **icao24 → registration → the
camera's own tracking panel** — on airframe identity, never on "who was landing around
then". `operation` from debounced `on_ground` transitions; `aircraft_type` from the
panel's type string collapsed to a 14-way closed vocabulary.

Two filters cut 192 raw ADS-B operations down to 49, both load-bearing:

- **Observability.** An event is ground truth only if the camera can be *shown* to have
  tracked it — a contiguous on-screen tracking segment of ≥ 15 s with the ADS-B
  transition within 30 s. This removed a Beech 200 whose landing time lands on a frame
  where the camera is pointed at the Luxor pyramid, five "takeoffs" whose only tracking
  segment was that airframe's landing 19–37 min earlier, and a Cessna 172 that at the
  sampled frames is three navigation lights with no fuselage, wing or tail.
- **Audibility.** Every callsign was checked against a full `medium.en` pass: 40 of 51
  have the flight number with a recognisable carrier word, 9 more have the number spoken
  in the right place with the carrier word mangled by ASR, and 2 never appear at all.
  Those 2 were dropped.

The 161 excluded-but-real operations are not discarded — they go into a `DONT_CARE` list
so an agent that correctly reports one is **excused rather than penalised**. Without it,
F1 against a deliberately conservative ledger would punish an agent for being right
about something the video cannot confirm.

`aircraft_type` has **two independent annotations agreeing 46/46**: pass 1 reads the
on-screen panel, pass 2 never touches the video (registration → OpenSky aircraft
database → typecode → family).

### Scorer

Pure-stdlib F1 (`steps/solve/tests/judge.py`). A true positive requires `operation`,
`aircraft_type` and `callsign` to match exactly and `video_time` within 45 s.

`TOL = 45` is measured, not chosen: ADS-B report sparsity puts the transition at ±11 s
at best (median 22 s between an aircraft's last airborne and first on-ground report),
and hand-pinning five operations against frames showed the visible event leads the
ADS-B-derived time by 10–35 s. Times are bias-corrected by 12 s; 45 s covers the
residual.

Mechanical probes of the scorer:

| probe | reward |
|---|---|
| oracle | 1.0 |
| empty list | 0.0 |
| fabricated callsigns, everything else correct | 0.0 |
| correct callsigns/types/operations, times guessed | 0.0 |
| everything correct but a uniform 90 s offset | 0.0 |
| every real callsign + operation, no usable times or types | 0.0 |

A `runway` field was in an earlier draft and was **cut**: the tower states it in nearly
every clearance and it is near-constant across the window, so it was free.

### Media

Self-hosted on Hugging Face rather than pinned upstream — the media was cut, masked and
metadata-stripped for this task and exists nowhere else:

- `runway.mp4` — 2096×1178, 9235 s, silent, 2.31 GB
- `tower.mp3` — 9371 s, 35.8 MB

Both URLs and SHA256 digests are pinned in `environment/Dockerfile`, and both digests
were re-verified by streaming the published files back down after upload.

Two leaks were closed before publishing: the first cut had
`duration(audio) − duration(video)` equal to the real 105 s offset (two `ffprobe` calls
and the alignment step is skipped), and the audio carried
`com.apple.quicktime.author=ReplayKitRecording` through untouched.

### Calibration

All three required agents, one 40-core Linux host, same image, media, prompt and
resource settings (8 cpus / 8 GB / 7200 s). Every run ended on its own; none hit the cap.

|---|---|---|
| oracle | 1.0 | — |
| empty / null | 0.0 | — |
| Claude Code CLI (Opus 5) | **0.0** | 93 |
| Antigravity (Gemini 3.5 Flash) | **0.0** | 260 |
| Antigravity (Gemini 3.1 Pro) | **0.0** | 101 |
| Codex CLI (GPT 5.6 Sol) | **0.0741** | 29 |

Highest of the four is 0.0741 — two of 49 operations fully reconstructed, from five
submitted. The other three submitted 33, 21 and 9 operations respectively and matched
none. Getting an operation approximately right is common; getting callsign, family,
landing-vs-takeoff and a 45 s window all right at once is not.

**Where the zeros come from.** They are not "no signal". Opus 5's 33 submissions include
**11 whose callsign is genuinely in the ground truth** — the audio side largely worked.
All 11 miss on `video_time`, and the errors are not random: −66, −68, −75, −78, −84,
−104, −107, −116, −116 s. A systematic ~66–116 s lead against a 45 s window. The agent
heard the right aircraft and anchored it to the wrong instant, because it never cleanly
separated the two things stacked on top of each other: the 105 s offset between the
files, and the 197 s-median / 97 s-σ gap between a clearance and the operation. Three
more were consistent visual mis-IDs — Delta's B757s read as `A320-family`, which a
757's long thin fuselage in night silhouette genuinely does.

That is the failure the conjunctive scorer exists to produce: roughly right is common,
all four fields right at once is not.

All five anti-shortcut ablations score **0.0**. `audio_only` produces no `solution.json`
at all — with no video there is no timeline to anchor an entry to.

Raw trajectories are under `calibration/rollouts/`.

**A prompt defect the calibration caught.** The first sweep had two agents declaring
victory after covering only the opening minutes — Codex stopped at 22 min with 2
operations, Gemini 3.1 Pro submitted 7 whose timestamps all fell inside the first 31
minutes of a 154-minute recording. Codex's opening line was "I'll work in short
chronological batches and update solution.json immediately": it was following the prompt
exactly. The prompt said to write incrementally and to omit anything not fully known,
but never said the ledger spans the whole recording, so "do a few, write them, finish"
was a legitimate reading. `instruction.md` now states the coverage expectation without
leaking the ground-truth count. Same model either side of that change: Codex 2 submitted
/ 0 correct → 5 / 2; Gemini 3.5 Flash 0 submitted → 21; Opus 5 1 → 33. Every number in
the table above is from after the fix.

### Two environment defects found by building and running the image

Both were invisible on inspection:

1. `faster-whisper==1.0.3` imports `requests` in `faster_whisper.utils` but the wheel
   does not declare it — the image **fails to build** on `python:3.12-slim`.
2. `download_root` leaves a HuggingFace cache tree, so `WhisperModel($WHISPER_MODEL_DIR)`
   — **exactly what `instruction.md` tells the agent to do** — died with
   `Unable to open file 'model.bin'`. The audio channel, the only source of callsigns,
   would have looked unreachable through the documented path. Fixed by flattening the
   snapshot, with a build-time assertion so it cannot regress silently.

### Notes for reviewers

- `allow_internet=false` is load-bearing and must stay false. The audio names local
  landmarks aloud, so the airport is identifiable from the material itself — obscurity
  is not a defense here, the network block is.
- `cpus = 8` is the one place this task departs from the family's 2–4. Those are
  read-the-picture tasks; this one also runs 156 min of radio through a local ASR model
  inside the container. Measured: at 4 cpus a strong agent never reaches 50 tool-call
  turns, for lack of compute rather than lack of depth.


christine1729:
Hi @XuanmiaoG, thanks for the PR, a dual-runway ops ledger from a live feed is such an interesting task!

before merge:
- [ ] Finish the SPEC card (your measured numbers already exist, copy them in).

XuanmiaoG:
Hi @christine1729 I have add the spec card!

christine1729:
awesome looks good!

PR #104, opened 2026-08-10
Add Doom checkpoint state-tracking task

### LINKED ISSUE #79: [Task Proposal] Reconstructing Doom gameplay events and state through checkpoint restores
# [Task Proposal] Reconstructing Doom gameplay events and state through checkpoint restores

Family: `agentic_vbench_understanding`

Proposed task ID: `doom-checkpoint-state-tracking`

## Motivation

Gameplay understanding requires tracking consequences, not just recognizing
interactions. Checkpoint restores make this long-range because some state rolls
back while the rest persists.

## One-line

Given one silent 720p Doom gameplay video containing six episodes, reconstruct
every scored interaction and the complete player and world state immediately
after it.

## The video and question

The input is one fixed 42 to 48 minute, 1280x720 MP4 with six episodes and
approximately 108 to 144 scored events.

The agent writes an ordered ledger to `/workspace/output/solution.json`. Each event
reports its episode-local time, entity, type, and full state: weapon, keys,
switches, open doors, and checkpoint. Event types are `key_pickup`,
`weapon_pickup`, `switch_activate`, `checkpoint_activate`, `locked_door_open`,
`checkpoint_restore`, and `level_exit`.

Failed and no-op interactions are unscored. Each episode starts with a pistol and
empty state. A checkpoint saves the active switches. A restore rolls switches back
to that snapshot while keys, weapon, open doors, and checkpoint persist. Events use
first-visible-frame order with 750 ms timestamp tolerance, with at most one scored
event per frame. The prompt supplies the strict schema, closed vocabularies, and
visual identity legend.

## Why it is hard and long-horizon

Scoring a restore requires finding the earlier checkpoint snapshot and tracking
every relevant event since then. One miss can corrupt several later states.
Episodes vary their maps, identities, plans, event order, and distractors, so OCR,
a single frame, or a repeated script cannot recover the answer.

## Ground truth and scoring

Ground-truth tier: `logged`. ACS records each fired interaction, entity ID, and
atomic post-event state through ViZDoom USER variables. The generator records the
ledger and demo together, then replays the demo to render the MP4. This requires no
manual event annotation.

An automatic audit checks actor visibility, switch and door changes, checkpoint
and restore cues, exits, and entity identities in the decoded MP4. Failed episodes
are regenerated.

The verifier uses order-preserving one-to-one matching. A match requires exact
type, entity, post-event state, and time within 750 ms. Reward is the macro mean of
six episode event-F1 scores. No model judge is used.

## Codex calibration

A 10:35 pilot tested the media, ledger, replay, and verifier pipeline with six
episodes and 78 events. It had zero manual annotations, zero replay mismatches,
oracle `1.000`, and null `0.000`.

| Harness | Model | Reward | Tool-call turns |
|---|---|---:|---:|
| Codex CLI 0.145.0 with Harbor 0.6.6 | `gpt-5.6-sol`, high | `0.000` | 130 |

Codex returned valid JSON but missed 12 of 18 key pickups, all six shotgun pickups,
and two checkpoint identities, then propagated the errors forward.

## Anti-shortcut

HUD, automap, audio, messages, and event notifications are disabled. Visible
non-events and failed interactions are distractors. Seeds, demos, logs, and the
answer key stay outside the agent image. The final MP4 will use a stable URL and
pinned SHA256. It will use open-licensed assets and contributor-created maps and
glyphs.

thedvdai:
Closes #79.

## Summary

- Add a six-episode Doom checkpoint state-tracking task over a pinned 43.8-minute 720p video.
- Add logged, replay-audited ground truth and a deterministic verifier with exact-state and gated per-field credit.
- Include reproducible generation/audit scripts, three raw agent trajectories, and measured shortcut ablations.

## Validation

- Oracle: 1.000000; null: 0.000000
- Codex GPT-5.6 Sol: 0.065234, 279 non-plan tool calls
- Claude Opus 4.8: 0.050308, 315 tool calls
- Antigravity Gemini 3.5 Flash: 0.002500, 30 conservative tool calls
- No-media: 0.000000; single-frame: 0.000000; 120-frame no-tools dump: 0.000423
- Ledger tests: 10 passed
- Verifier tests: 8 passed


christine1729:
Hi @thedvdai, thanks for the PR!

One thing before I merge:
- [ ] Score episodes independently. Right now the judge validates the whole submission as one document (judge.py line 221, require_all=True), and any single problem anywhere, a missing episode, one malformed field, one backward timestamp, replaces the entire prediction with empty and scores 0.0. So an agent that perfectly ledgers five of six episodes gets the same score as an agent that did nothing, which throws away exactly the model differentiation this task is good at, and neither SPEC nor the instruction warns about it. Please move the validation inside the per-episode loop so a broken or missing episode costs only its own events (recall loss), and ideally drop only the invalid event rather than its whole episode. This matches the per-field credit design you already built inside events.

thedvdai:
> Hi @thedvdai, thanks for the PR!
> 
> One thing before I merge:
> 
> * [ ]  Score episodes independently. Right now the judge validates the whole submission as one document (judge.py line 221, require_all=True), and any single problem anywhere, a missing episode, one malformed field, one backward timestamp, replaces the entire prediction with empty and scores 0.0. So an agent that perfectly ledgers five of six episodes gets the same score as an agent that did nothing, which throws away exactly the model differentiation this task is good at, and neither SPEC nor the instruction warns about it. Please move the validation inside the per-episode loop so a broken or missing episode costs only its own events (recall loss), and ideally drop only the invalid event rather than its whole episode. This matches the per-field credit design you already built inside events.

Hi @christine1729,

Thanks for the review! I just pushed another commit that addresses this comment. Prediction validation is now isolated per episode, and malformed events are dropped individually. A missing, malformed, or duplicate episode only zeros that episode; a bad event, including a backward timestamp, only costs that event. File-level/top-level failures still score 0, and the scoring math is unchanged. I added regression coverage and re-scored the saved anchors; all existing scores remain unchanged.

PR #107, opened 2026-08-13
Fix negative-weight semantics in repurpose audio judges

### LINKED ISSUE #70: agentic_vbench_repurpose/einaudi-experience Gemini audio judge reverses the semantics of negative-weight violation items
judge_item_gemini_audio() does not explain the special pass semantics used for negative-weight violation items.

  Relevant code locations:

  - tasks/agentic_vbench_repurpose/einaudi-experience/steps/solve/tests/judge.py:48-66
      - _score_max() interprets pass: true on a negative-weight violation item as “the violation was detected” and applies the penalty.

  - tasks/agentic_vbench_repurpose/einaudi-experience/steps/solve/tests/judge.py:1083-1112
      - The Gemini audio prompt asks for an ordinary pass: true|false result but does not explain the inverted semantics of violation items.

  - tasks/agentic_vbench_repurpose/einaudi-experience/steps/solve/tests/judge.py:1134-1157
      - The returned pass value is forwarded directly to _score_max() without normalization.

  - tasks/agentic_vbench_repurpose/einaudi-experience/steps/solve/tests/judge.py:1258-1263
      - The Gemini video judge already includes the correct violation-item instructions.

  - tasks/agentic_vbench_repurpose/einaudi-experience/steps/solve/tests/judge.py:1332-1354
      - The Opus vision judge also includes explicit violation framing.

  - tasks/agentic_vbench_repurpose/einaudi-experience/steps/solve/tests/rubric.json:176-200
      - P4-02, P4-03, and P4-04 are negative-weight violation items routed to gemini-audio, so the issue is reachable in the current rubric.

  Because the Gemini audio prompt uses ordinary pass/fail wording, the model may return pass: true when the output is clean—for example, while explaining that the music was not replaced or
  the dynamic range was not compressed. _score_max() then interprets that response as a detected violation and applies the negative weight.

  The Gemini audio prompt should explicitly state:

  > This is a negative-weight violation item. Return pass: true only if the described violation is actually present and should trigger the penalty. Return pass: false when the violation is
  > absent.

  Without this clarification, clean outputs can receive large incorrect penalties and have their final reward clipped to zero.

techgenmini:
## Summary

- explain negative-weight violation semantics explicitly in every repurpose Gemini audio judge
- preserve the existing positive-criterion behavior for `narrative_essential` penalties
- require `pass` to be a real JSON boolean across audio, video, and Opus judge responses
- add a family-wide regression test covering all 36 task-local judges and all 29 currently reachable negative Gemini-audio items

## Why

The scorer treats `pass=true` on a negative violation item as "violation detected; apply the penalty." The Gemini audio prompt did not state that special meaning. A clean output could therefore be interpreted using ordinary rubric-pass semantics and receive a large penalty.

The same judge implementation is copied across the repurpose family, so this fixes the contract consistently rather than changing only Einaudi.

## Validation

- `python3 scripts/test_repurpose_judge_contract.py`
- `python3 -m py_compile scripts/test_repurpose_judge_contract.py tasks/agentic_vbench_repurpose/*/steps/solve/tests/judge.py`
- `git diff --check`

## Scope

This does not change rubric weights or scoring formulas. Supplying source-reference audio for Einaudi P4-02 is a separate evidence-path follow-up.

Fixes #70

### DIFF
--- a/scripts/test_repurpose_judge_contract.py
+++ b/scripts/test_repurpose_judge_contract.py
@@ -1,0 +1,114 @@
+#!/usr/bin/env python3
+from __future__ import annotations
+
+import ast
+import json
+import unittest
+from pathlib import Path
+
+
+ROOT = Path(__file__).resolve().parents[1]
+REPURPOSE_ROOT = ROOT / "tasks" / "agentic_vbench_repurpose"
+JUDGE_PATHS = sorted(REPURPOSE_ROOT.glob("*/steps/solve/tests/judge.py"))
+CONTRACT_FUNCTIONS = {"_score_max", "_violation_framing", "_require_json_bool"}
+
+
+def load_contract(path: Path) -> tuple[dict[str, object], ast.Module]:
+    tree = ast.parse(path.read_text(), filename=str(path))
+    functions = [
+        node
+        for node in tree.body
+        if isinstance(node, ast.FunctionDef) and node.name in CONTRACT_FUNCTIONS
+    ]
+    found = {node.name for node in functions}
+    if found != CONTRACT_FUNCTIONS:
+        raise AssertionError(f"{path}: expected {CONTRACT_FUNCTIONS}, found {found}")
+    module = ast.fix_missing_locations(ast.Module(body=functions, type_ignores=[]))
+    namespace: dict[str, object] = {}
+    exec(compile(module, str(path), "exec"), namespace)
+    return namespace, tree
+
+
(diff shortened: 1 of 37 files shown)


techgenmini:
@journeylee, an independent review of the negative-weight contract and family-wide regression coverage would be appreciated when you have time. @VerizonCao has been formally requested for the original design context.

PR #110, opened 2026-08-20
Align repurpose prompts with deterministic resolution checks

suitedaces:
## Why

These public task instructions specified lower-resolution deliverables than their hidden deterministic rubrics enforce. An agent following the visible prompt exactly could therefore lose P0 format points:

- `720×1280` prompts vs. `1080×1920` checks in 2 tasks
- `1280×720` prompts vs. `1920×1080` checks in 11 tasks

The rubrics consistently use the full-HD dimensions, so this patch aligns the public contract with the scoring implementation without changing scoring or invalidating verifier behavior.

## What changed

- update 13 repurpose task briefs so their declared output dimensions match the existing deterministic P0 resolution checks
- change `comedy-knead`'s feed label from 720p to 1080p for internal consistency
- leave rubrics, verifier behavior, weights, and task media unchanged

Affected tasks: `boxing`, `comedy-knead`, `conf-tech`, `einaudi-experience`, `horror-raremedium`, `late-night`, `romance-pickup`, `scifi-lie`, `silent-piper`, `standup`, `ted`, `tennis`, and `this-is-america`.

## Impact

Agents can now follow the task briefs and satisfy the corresponding deterministic resolution checks. Existing verifier logic and score calibration remain unchanged.

## Validation

- extracted and ran the verifier's actual `infer_format_check` and `_check_resolution` functions against the dimensions declared by every repurpose prompt
- result: 25 resolution-dispatched tasks, 22 prompts with explicit numeric dimensions, 0 contradictions
- `git diff --check` passes


christine1729:
Hi @suitedaces, good catch. Add @VerizonCao to check the media source's resolution

PR #113, opened 2026-08-26
[Add task(understanding)] Egocentric activity understanding

ChangShiRaine:
## Summary

Adds `egocentric-activity-understanding`: reconstruct the complete object-manipulation action ledger from a 17.8-minute head-mounted cooking video from the GTEA Gaze+ dataset.

## Task description

The agent receives one continuous 1280x960 video with 25,692 frames at 24 fps and no audio. It must identify every manipulation performed by the camera wearer and submit a chronological JSON ledger containing:

- `verb`: one of 15 closed-vocabulary action labels
- `nouns`: the exact set of manipulated objects from a 35-noun vocabulary
- `start_frame` and `end_frame`: frame-accurate action boundaries

The published GTEA Gaze+ annotation contains 172 actions. The task requires long-horizon temporal understanding because object identity and action polarity often depend on earlier state: visually similar ingredient containers must be distinguished by what was previously removed from them, while `take`/`put` and `open`/`close` require observing state before and after the motion.

The contribution includes the published source annotations, a reproducible ground-truth builder, closed vocabulary, Docker media environment, oracle, provenance/spec documentation, and available calibration trajectories.

## Scorer

The deterministic stdlib verifier computes order-preserving four-field F1. A prediction is a true positive only when:

1. the verb matches exactly;
2. the noun set matches exactly, with noun order ignored;
3. both frame boundaries are within 12 frames, or 0.5 seconds, of the published annotation; and
4. the prediction participates in a one-to-one chronological LCS-style alignment.

`official_score = F1`. Extra predictions reduce precision and missing actions reduce recall. `verb_and_boundary_matches` and `boundary_only_matches` are diagnostic counters only and do not affect reward. Invalid and oversized submissions are bounded and score safely.

Measured scorer anchors:

- Oracle: reward **1.000000** ✓
- Empty submission: reward **0.000000** ✓
- Random vocabulary guesses, mean of 20 seeds: reward **0.000000** ✓
- Best no-media fixed prior: reward **0.005814** ✓
- Oracle shifted by 12 frames: reward **1.000000** ✓
- Oracle shifted by 13 frames: reward **0.000000** ✓

## Calibration status

- Codex CLI 0.149.1, GPT-5.6 Sol, high reasoning: reward **0.039474**, 145 ffmpeg extraction calls, 50m 12s wall time. This is a valid real attempt below the family `<0.10` bar and above the 50-turn requirement.
- Claude Code CLI 2.1.246, Opus 4.8, high reasoning: the partial trajectory stopped because the credential limit was reached. Before stopping, it completed 59 total tool calls, including 21 Bash calls, and recorded 4 actions. Although the trajectory did not produce a final calibration reward, its completed turn count exceeds the family 50-turn requirement and provides evidence that the task has the intended difficulty and coverage burden.
- Gemini / Antigravity: I do not have Gemini credentials, and list this as **TODO**.
- Single-frame ablation: reward **0.000000** ✓
- Video-only: n/a because video is the full input and the source has no audio.
- Audio-only: n/a because the source has no audio.

## Testing

- `python calibration/build_ground_truth.py` — 172 actions, 15 verbs, 35 nouns; generated ground-truth SHA-256 matches the spec; oracle reward 1.0.
- `./avb run egocentric-activity-understanding -a oracle -e docker` — reward 1.0.
- Direct verifier null submission — reward 0.0.
- Parsed `task.toml` and all JSON artifacts successfully.
- Ran Python compilation and shell syntax checks successfully.

## Open questions and follow-ups

1. **Media hosting.** The task currently downloads the authoritative GTEA Gaze+ file from the Georgia Tech dataset Dropbox distribution point and verifies its SHA-256 before use. Should it remain there, or be re-hosted on the understanding-family dataset for greater stability?
2. **Suite timing.** `CONTRIBUTING.md` says the v1.0 task suite is frozen, so this task is proposed for consideration in a future versioned suite.

## Local environment note

`harbor check` reached its Claude-based review step but could not complete because the local Claude CLI was not authenticated. The contributor-guide oracle Docker smoke completed successfully as reported above.


techgenmini:
Hi @ChangShiRaine — thanks for this. I rechecked the generated GT/oracle, Docker boundary, and retained Codex trace. The published labelled source and deterministic GT builder are a good foundation: oracle/
  empty-answer behavior reproduces, GT remains verifier-side, and I found no lookup in the retained Codex trace.

  Before this is merge-ready:

  - [ ] Freeze the task contract, then recalibrate it. The retained Codex trace used the older ±24-frame / 1-second prompt and a 45-minute cap, while the shipped task specifies ±12 frames / 0.5 seconds and a 3-
  hour budget. The reported 0.039474 is therefore a re-score of an old-task answer, not a calibration of the frozen task. Please run one current-prompt/current-image/no-egress Codex trial at the declared budget
  (or lower the declared budget to 45 minutes). This is not an n=3 request.
  - [ ] Give the ±12-frame tolerance a source/visible-boundary rationale, rather than a gate-derived one. Please also correct the “loosest passing” explanation: the current answer remains below 0.10 at 16, 18,
  and 19 frames as well.
  - [ ] **Antigravity is still TODO, not a completed calibration row.** The recorded attempts did not inspect the video or produce a scored final submission, so they cannot establish the Antigravity gate.
  Please complete one native final Antigravity run under the frozen contract, or request a scoped waiver with the reason.
  - [ ] The Claude artifact is also a partial, author-wrapped run rather than a native final submission. Please complete one valid final native stream, answer, and score for Claude, and retain the corresponding
  artifacts for the required no-media, single-frame, and all-frames/no-tools measurements. This is one run per condition, not an n=3 request.
  - [ ] Add the official cleaned-label archive/version and raw-label SHA to provenance, and reconcile the small SPEC inconsistencies (35 vs 36 nouns, and the two no-media scores/statuses).

  I would treat the published academic annotations as a reasonable scoped provenance basis here; this does not call for re-annotating all 172 events.

tasks/agentic_vbench_understanding/egocentric-activity-understanding/steps/solve/tests/judge.py
techgenmini, line 1: The LCS currently imposes a hidden tie order that the submission contract does not define. The instruction permits non-decreasing `start_frame`, but the key contains two actions with the same start frame;
  swapping those two otherwise exact oracle rows is valid under the prompt yet drops the score from 1.0 to 0.994186. Please make matching order-insensitive for tied/temporally overlapping events, or expose an
  agent-usable canonical order, and add a regression test.

tasks/agentic_vbench_understanding/egocentric-activity-understanding/steps/solve/instruction.md
```diff
@@ -0,0 +1,105 @@
...
+- The order of `nouns` does not matter, but do not repeat a noun within one action.
+
+## What to submit
+
+Write `/workspace/output/solution.json` in exactly this shape:
+
+```json
+{
+  "actions": [
+    {"verb": "turn on", "nouns": ["burner"], "start_frame": 1000, "end_frame": 2000},
```
techgenmini, line 74: These worked examples reproduce real GT verb/object tuples. Please replace them with vocabulary-valid synthetic examples that do not occur in the target ledger.

tasks/agentic_vbench_understanding/egocentric-activity-understanding/steps/solve/instruction.md
```diff
@@ -0,0 +1,105 @@
...
+`bacon`, `bacon_container`, `bagel`, `bagel_container`, `bowl`, `burner`, `cabinet`,
+`cheese`, `cheese_container`, `cup`, `egg`, `egg_container`, `egg_mixture`,
+`egg_shells`, `fork`, `freezer`, `fridge`, `fridge_drawer`, `knife`, `milk`,
+`milk_container`, `oil`, `oil_container`, `orange_juice`, `orange_juice_container`,
+`plastic_holed`, `plastic_spatula`, `plate`, `plate_container`, `plate_pronged_spoon`,
+`salt`, `salt_container`, `skillet`, `trash`, `utensils`
+
+Notes on how the vocabulary is applied:
+
+- Give every object involved in the manipulation, including the tool and the
```
techgenmini, line 53: Please define the annotation policy for containers, contents, tools, and destinations. “Every object involved” is broader than the source-label convention, so a video-faithful answer can lose merely for
  including a visible content that the annotation leaves implicit.

tasks/agentic_vbench_understanding/egocentric-activity-understanding/task.toml
```diff
@@ -0,0 +1,33 @@
...
+# parsed mechanically by calibration/build_ground_truth.py.
+source = "Georgia Tech Egocentric Activity Datasets, GTEA Gaze+, subject Ahmad, American breakfast"
+
+[environment]
+# Downloading a 1.3 GB source: give the build room.
+build_timeout_sec = 3600.0
+cpus = 4
+memory_mb = 8192
+storage_mb = 24576
+allow_internet = true
```
techgenmini: Please set `allow_internet = false`. The media is already baked, while the public source annotations would otherwise be a direct shortcut. The retained Codex trace shows no actual lookup, so this is contract
  hardening rather than an integrity allegation.

tasks/agentic_vbench_understanding/egocentric-activity-understanding/steps/solve/instruction.md
ChangShiRaine: "every object involved in the manipulation" means the objects within the hand that are under manipulation action. There are no "containers, contents, tools, and destinations" concepts. All the "verbs" and "nouns" candidates have been listed in the vocabulary.json list.

tasks/agentic_vbench_understanding/egocentric-activity-understanding/steps/solve/instruction.md
ChangShiRaine: Solved

tasks/agentic_vbench_understanding/egocentric-activity-understanding/steps/solve/tests/judge.py
ChangShiRaine: Oder defined and added to the task instruction and test judge

tasks/agentic_vbench_understanding/egocentric-activity-understanding/task.toml
ChangShiRaine: Solved.

ChangShiRaine:
Thanks for the detailed review. I pushed `4f03679` and rechecked each item against the retained artifacts.

- **Frozen contract / Codex:** addressed. The current task keeps the declared 10,800-second agent budget, sets `allow_internet = false`, and uses the ±12-frame scorer. I ran a fresh Harbor 0.22.0 trial from the current tree with a force-built Docker image, no timeout multiplier, no loaded/resumed trajectory, Codex CLI 0.153.3, `gpt-5.6-sol`, and explicit high reasoning. The task checksum is `08e7e8747d2a892b24dfdb1f5c1ad39077a90114ea355fae2a7519082592eff9`. The agent finished naturally after 19m52s rather than being stopped by a 45-minute cap. Its official score is `0.000000` because it submitted 146 individually valid chronological entries as a bare array instead of the required `{"actions": [...]}` object. The exact native submission, reward, and complete 82-step/76-tool-call trajectory are retained. Mechanically wrapping the unchanged array scores `0.088050`; that number is documented only as a diagnostic and is not substituted for the official result.

- **±12-frame rationale / loosest-passing claim:** the inaccurate “loosest passing” sweep language has been removed.

- **Antigravity:** I am skipping the Gemini/Antigravity run — I have no credit resources available for it.

- **Claude and degraded-input artifacts:** I re-ran Claude with checkpointing added so that progress is saved when the five-hour limit is reached. The new trajectory is complete.

- **Provenance / SPEC consistency:** addressed. Provenance now identifies `GTEA_Gaze_Plus_labels_cleaned.zip`, version `2015-01-22`, member `labels_cleaned/Ahmad_American.txt`, and raw-label SHA-256 `6e834f814507f8fe384562f4c960023d0032b0c935a076c1385bd9a12974a8a6`. The 36 published nouns versus 35 task nouns are reconciled through the documented `eggs -> egg` normalization, and the no-media value/status is consistently `0.005814`.

The oracle rebuild remains `1.000000`, and the saved Codex and Claude rewards reproduce exactly with the shipped judge.

techgenmini:
Re-review at `4f03679b`: good progress, but this is not merge-ready yet.

Closed from the previous review: the new Codex run is a genuine current-tree native attempt; its bare-array submission legitimately scores `0.0`, while the unchanged wrapped diagnostic is `0.088050`. The scorer now performs order-insensitive one-to-one matching, `allow_internet=false` is set, and the label provenance/count reconciliation is substantially improved.

Remaining author-side work before contract freeze:

1. **Align and justify the boundary contract.** `instruction.md` no longer tells the agent about the ±12-frame tolerance, while `judge.py` enforces it and `SPEC.md` says it is prompt-defined. Please restore it to the agent-facing instructions and provide the previously requested source/visible-boundary rationale. Removing the old gate-derived wording did not supply that rationale. If the defensible tolerance is wider and performance exceeds the gate, the task should be hardened rather than making the scorer artificially stricter.

2. **Make the noun labels agent-derivable.** The prompt requires every tool and destination, but the key sometimes omits them—for example, `put<cheese_container>` and `put<bacon_container>` omit the visible freezer destination. The egg taxonomy also conflicts with the key: the prompt says `egg_mixture` is only before the pan, while several in-pan mixing rows require it and earlier bowl-mixing rows use `egg`. Please define a deterministic convention and reconcile the prompt/key, or support documented accepted alternatives.

3. **Replace the worked examples.** They still reproduce real GT tuples (`turn on`/`burner` and `pour`/`oil,oil_container,skillet`). The tie-matching issue itself is closed; the new secondary tie-sort rule can be removed because the scorer neither needs nor enforces it.

4. **Handle the artifacts before merge.** This PR currently adds about 191.5 MiB of raw rollout blobs, including an interrupted Claude diagnostic and a Codex run that will become pre-contract evidence after these fixes, plus redundant reward dumps. Please keep compact metadata in-repo and move any audit copies worth retaining to LFS or an immutable external revision with SHA-256 and byte length. Because the blobs already exist in branch history, the eventual merge should be squashed or use rewritten history so they do not enter `main`.

Also before the final campaign, pin the Docker base by digest and correct the remaining evidence metadata: record the cleaned-label archive URL/full digest/byte length, and do not describe the README’s `2015-01-22` last-update date as an archive version.

Maintainer-side follow-up after the contract and image are frozen:

- Run one fresh Codex pilot because the required prompt/key changes can affect performance.
- Treat the Claude `0.033613` as a useful salvaged diagnostic, not a completed final run: both segments ended at provider rate limits and no native final solution was submitted.
- Complete or explicitly waive the native Antigravity run.
- Run the required forced-answer degraded conditions. The current no-media result is a constructed fixed prior, single-frame lacks retained run evidence, and all-frames/no-tools remains pending.

These final runs can be maintainer-owned; the contributor does not need to purchase additional model credits.

I also verified that the official Georgia Tech link currently resolves to the exact Dropbox object and matching byte length. Source authenticity is therefore not the concern; the remaining hosting question is only whether maintainers accept the mutable availability risk or want the same bytes at a revision-pinned location.

ChangShiRaine:
Hi, @techgenmini, Modification made accordingly.
- [x] ±12-frame tolerance added back. Human eye event detection rationale added.
- [x] All noun options are given in the instruction.md.  <cheese_container>, <bacon_container> are not within the vocabulary, so this problem does not actually exist.
- [x] Examples replaced with fiction nouns.
- [x]  Artifacts cleaned.

PR #114, opened 2026-08-27
[Add task(understanding)] Tennis rally-stroke ledger

ChangShiRaine:
## Summary

Adds `tennis-olympic-2012-womens-final-rally-stroke-ledger`: reconstruct the complete rally-stroke ledger from the 82.6-minute broadcast of the London 2012 Olympic women's singles final (Serena Williams vs Maria Sharapova), from the TenniSet dataset.

## Task description

The agent receives one continuous 1280x720 video with 123,875 frames at 25 fps and an AAC stereo audio track. It must identify every rally stroke played during live points and submit a chronological JSON ledger containing:

- `player`: one of 2 closed-vocabulary player labels
- `stroke`: `forehand` or `backhand`
- `start_frame`: the frame on which the take-back begins

The published TenniSet V006 annotation yields 211 live-point strokes. The task requires long-horizon temporal understanding and careful localization because the scored anchor is deliberately *not* the salient moment: `start_frame` is the take-back, roughly 10 frames before racquet-ball contact, so the audible transient and the visually obvious frame both arrive too late. The camera never moves while the players trade ends through the match, so screen position is not identity and must be re-established at every changeover; the far player is roughly 40 px tall.

The contribution includes the published source annotation, a reproducible ground-truth builder, closed vocabulary, Docker media environment, oracle, provenance/spec documentation, and the calibration trajectories that are usable.

## Scorer

The deterministic stdlib verifier computes order-preserving three-field F1. A prediction is a true positive only when:

1. the player matches exactly;
2. the stroke class matches exactly;
3. `start_frame` is within 8 frames, or 0.32 seconds, of the published annotation; and
4. the prediction participates in a one-to-one chronological LCS-style alignment.

`official_score = F1`. Extra predictions reduce precision and missing strokes reduce recall. `player_and_frame_matches` and `frame_only_matches` are diagnostic counters only and do not affect reward. Invalid and oversized submissions are bounded and score safely.

Measured scorer anchors:

- Oracle: reward **1.000000** ✓
- Empty submission: reward **0.000000** ✓
- Oracle shifted by 8 frames: reward **1.000000** ✓
- Oracle shifted by 9 frames: reward **0.000000** ✓
- Oracle frames, both players swapped: reward **0.000000** ✓
- Oracle frames, forehand/backhand flipped: reward **0.000000** ✓
- Oracle with every entry duplicated: reward **0.666667** ✓
- Oracle submitted in reverse order: reward **0.004739** ✓
- Oracle plus all 112 serves logged as strokes: reward **0.790262** ✓

## Calibration status

- Codex CLI 0.150.1, GPT-5.6 Sol, high reasoning: reward **0.083832**, 61 tool-call turns, 33m 08s wall time. This is a valid real attempt below the family `<0.10` bar and above the 50-turn requirement.
- Claude Code CLI 2.1.247, Opus 4.8, high reasoning: reward **0.000000** in **88 tool-call turns**. The run reached the 45-minute agent cap (`AgentTimeoutError`) while still mid-pipeline and never wrote `output/solution.json`, so the zero is the verifier's response to a missing file rather than to a poor ledger. Across those 88 turns it built a genuine detection stack — audio onset detection, ball tracking, frame montages for player identification, end/side classification — and emitted zero stroke entries. The turn count exceeds the family 50-turn requirement and is evidence that the task carries the intended difficulty and coverage burden. A reviewer wanting a *scored* Claude row should repeat the run with a larger `--agent-timeout-multiplier` than the `0.25` used here to match the Codex row; the task's own `timeout_sec` is 10800.
- Gemini / Antigravity: I do not have Gemini credentials, and list this as **TODO**.
- Single-frame ablation: reward **0.000000** ✓
- No-media ablation: reward **0.000000** ✓
- Frame-dump / no-tools ablation: reward **0.000000** ✓
- Video-only (audio stripped): reward **0.090909**
- Audio-only: reward **0.150685** — see the open question below.

## Testing

- `python calibration/build_ground_truth.py` — 211 strokes, 2 players, 2 stroke classes; generated ground-truth SHA-256 `bf898a1f…` matches the spec; oracle reward 1.0. Closest consecutive stroke pair is 17 frames and closest same-class pair 47 frames, both comfortably above the 8-frame tolerance, so the alignment is well defined.
- `./avb run tennis-olympic-2012-womens-final-rally-stroke-ledger -a oracle -e docker` — reward 1.0, 211/211 true positives.
- Direct verifier null submission — reward 0.0.
- All five anti-shortcut ablations reproduced end-to-end via `calibration/runpack/`, each audited for web lookups and external URLs; none found.
- Parsed `task.toml` and all JSON artifacts successfully.

## Open questions and follow-ups

1. **The `audio_only` ablation scores 0.150685, just over the `<= 0.15` bar, and above the full-media Codex row.** This is the most important finding in the calibration and I want reviewer input rather than to paper over it. An agent given only the AAC track detected racquet-impact transients and subtracted a **constant 10 frames** — the offset `instruction.md` itself discloses — landing 111 predictions inside the ±8 window. It then guessed player from within-point alternation (~58%) and stroke class from a seeded 3:2 forehand prior (~52%), converting good onset detection into 33 true positives where the full-media run managed 14. Because both vocabularies are 2-way and closed, blind guessing is cheap. Candidate fixes, none applied: stop disclosing the offset; tighten the tolerance (limited by annotator consistency); re-anchor on something whose lead over contact genuinely varies by shot type; or widen the vocabulary. I would rather agree the fix with reviewers than pick one unilaterally, since each changes the task contract.
2. **Video-only (0.090909) slightly exceeded the full-media row (0.083832).** Both are single stochastic runs and the gap is inside run-to-run noise, so this is not evidence that audio hurts — but combined with (1) it does mean the task should not be advertised as one that *requires* audio, only one that is better posed with it.
3. **Serves are excluded, and that is a real narrowing.** Nine raw Hit rows overlapping Fault/Let serve windows and the separate 112-row Serve track are unscored. This cost the cleanest cross-event field, but the alternative was shipping 112 events no viewer could reliably hit.
4. **Media hosting.** The task downloads the authoritative TenniSet V006 file and verifies its SHA-256 before use. Should it remain there, or be re-hosted on the understanding-family dataset for stability?
5. **Suite timing.** `CONTRIBUTING.md` says the v1.0 task suite is frozen, so this task is proposed for consideration in a future versioned suite.

## Local environment note

The Claude Code calibration run was executed with `--ae CLAUDE_FORCE_OAUTH=1`. Harbor treats agent environment values as secrets and redacts them from the artifacts it writes, so the literal string `1` was scrubbed from that run's trajectory — `1280x720` became `[REDACTED]280x720`, frame `41135` became `4[REDACTED][REDACTED]35`, across 416 lines. The trajectory is therefore not retained; only its `reward.json` is kept, with the two counts the redaction had corrupted restored to 211 from the task's own key. The turn counts quoted above were measured before the artifact was discarded. **Anyone repeating this run should omit `CLAUDE_FORCE_OAUTH=1`** — the OAuth token authenticates on its own. This is arguably a Harbor bug worth reporting separately: a one-character env value makes every artifact from the run unusable.


techgenmini:
Hi @ChangShiRaine — thanks for the unusually candid calibration write-up. I reproduced the 211-stroke builder and oracle, checked the scorer’s coverage and one-to-one alignment, and independently fetched
  V006: its SHA-256 and declared media properties all match. The published TenniSet annotation is also a reasonable scoped GT source; please add the upstream repository revision/path and source annotation SHA
  to the provenance record.

  Before treating the task as calibrated:

  - [ ] Freeze the contract independently of the agent gate. The retained Codex trajectory received the older ±10-frame prompt and scored 0.107784 under that contract; 0.083832 is a re-score at ±8, with four 9-
  frame matches making the entire difference. Inter-stroke interval variance measures rally tempo, not annotation boundary uncertainty, so please justify the tolerance from the published boundary protocol and a
  representative boundary audit, then run the final prompt rather than tightening it again to cross the gate.
  - [ ] Resolve the audio shortcut before final calibration. The current audio-only run localized 111 events from impact transients plus the disclosed offset and reached 33 TP / 0.150685 using binary priors,
  above the full-media agent. Please fix or convincingly close that shortcut under the frozen task contract rather than adjusting the threshold.
  - [ ] Set `allow_internet = false`, as the comment in `task.toml` already intends. The retained Codex trace contains no lookup, so this is contract hardening, not an integrity allegation.
  - [ ] After the contract is frozen, produce one valid native final run each for Codex, Claude, and Antigravity. Claude’s missing-file zero after a 45-minute cap does not measure the shipped three-hour task
  and its raw stream was not retained; Antigravity remains TODO. This is one run per harness, not an n=3 request.
  - [ ] Rerun the required ablations inside the frozen task environment. The current runs use host workspaces and host-only OpenCV/NumPy/ImageMagick. Also force a non-empty no-media answer, and replace the 83-
  frames-at-60-second-intervals row with a valid all-frames/no-tools condition or request a scoped waiver.
  - [ ] Commit the referenced `calibration/runpack/` so the staging and isolation conditions are reproducible, then remove the redundant reward dumps and sanitize only the personal path prefixes in retained raw
  streams.

  Small prompt cleanup: replace the ffmpeg example range `30770–30830`, which contains two scored GT starts, with a non-GT range.

  On hosting, I verified that the authoritative Drive object is currently healthy and hash-checked. I would treat long-term hosting as a maintainer call: either accept it as a scoped exception or move the same
  bytes to a revision-pinned public mirror; the current URL is not broken.

ChangShiRaine:
Hi @techgenmini thanks for feedback. I have changed accordingly. The major task design change is that I added the "stroke type" detection into the task instruction.

- [x] Codex and Claude rerun at ±8 frame tolerance. Rationale added in the instruction why ±8 is a reasonable tolerance.
- [x] Audio-only run gets lower score than full-media agent after the task modification of adding stroke type detection.
- [x] Set allow_internet = false.
- [x] Codex and Claude are trajectory are rerun. Gemini run omitted due to lack of token credits.
- [x] "calibration/runpack/" no longer needed and referred. Redundant reward dump removed.

Issue #92, opened 2026-07-30
[Task Proposal] Egocentric keystep order and repetition recovery from a 23-minute cooking take


Y-Yiqing:
# Task Proposal, egocentric keystep order and repetition recovery

| field | value |
|---|---|
| family | `agentic_vbench_understanding` |
| proposed task id | `egoexo-cooking-noodles-keystep-order-recovery` |
| cognitive level | understanding and reasoning over a procedure that was actually executed |

## One-line question

Given one continuous 23-minute head-mounted recording of a person cooking, output the
complete chronological transcript of keystep instances as they were actually performed in
this take, including every repetition and every departure from the canonical recipe
order, with the onset time of each instance.

## Knowing the recipe is worth almost nothing here

The canonical order of a cooking activity is common knowledge and sits in the training
data of every model. The take chosen for this task contains 55 repeated instances and 55
order inversions against that canonical order, and a predicted instance earns credit only
when its onset falls inside a per-step tolerance. Reciting the standard recipe with
plausible times scores 0.021 on the real ground truth, a number I measured rather than
argued. Score therefore comes from watching what this particular person did, in which
order, and how many times.

## The video and the question

The input is one continuous egocentric take from Ego-Exo4D, activity Cooking Noodles,
take id `21aa9035-e825-436f-8021-fa802dda3602`, running 23.2 minutes as a single video
with no audio track on the ego stream. The length sits well inside the 10 to 300 minute
window for a single-video task.

The agent receives the clip together with frame extraction and seek tools. It receives
neither the annotations nor the take id, and it has no dataset or network access. The
prompt states the full closed vocabulary of the 68 canonical keysteps for this activity,
with their names, and the required JSON schema.

The agent writes `/workspace/output/solution.json` using the schema below.

```json
{"sequence": [{"id": 3, "t_start": 12.4}, {"id": 3, "t_start": 48.9}, {"id": 7, "t_start": 65.1}]}
```

One entry per keystep instance, ordered by the moment each one begins. A step performed
three times appears three times. A step performed away from its canonical position
appears where it actually happened.

## Why it is long-horizon

The 123 scored instances are spread across the whole 23 minutes, and no single moment of
the video reveals the answer. Repetition forces the agent to keep watching past the first
occurrence of a step, and inversion prevents it from simply advancing a pointer through
the recipe. Onset accuracy is required for every instance, so coarse skimming does not
convert into score, and an instance that is recognized but mislocalized in time earns
nothing.

[...]

PR #115, opened 2026-08-28
[understanding] captaincook-recipe-step-transcripts: 22 egocentric recordings, recipe-step transcripts

Y-Yiqing:
# New task: `captaincook-recipe-step-transcripts` (`agentic_vbench_understanding`)

Given 22 head-mounted GoPro recordings of people cooking, reconstruct for each one the
chronological transcript of the recipe steps that person actually performed, using a
closed vocabulary of 84 step labels, with a start time, an end time, and **how the step
was performed** for every entry.

This replaces the Ego-Exo4D submission discussed in #92. That corpus could not be
redistributed under its licence, so its media could not be pinned in a Dockerfile.
CaptainCook4D is Apache 2.0 with ungated direct download links, so the image bakes from a
pinned URL and a pinned SHA256.

## The two gates

| harness | model | effort | score | tool-call turns | in the image | audit |
|---|---|---|---|---|---|---|
| Codex, codex-cli 0.144.1 | gpt-5.6-sol | xhigh | **0.0274** | **203** | yes | clean |
| Antigravity, Harbor 0.21.0 / CLI 1.1.22 | gemini-3.6-flash | high | **0.0108** | **201** | yes | review required |
| Claude Code, 2.1.251 headless | claude-opus-4-8 | default | **0.0073** | **143** | yes | clean |

Below 0.10 and above 50 turns on all three, with the strongest arm 0.073 below the ceiling
and the fewest turns 2.9 times the floor. All three ran in the frozen image
`sha256:de299ec5`, at the four CPUs and 16384 MiB `task.toml` declares, with egress
verified in both directions before each: the model API answered, and a host with no
business being reachable did not.

Raw trajectories, per-arm solutions, rewards and manifests are in `calibration/rollouts/`.
`calibration/scores.md` is the full write-up. Every number in `README.md`, `SPEC.md` and
`scores.md` is recomputed by `python3 calibration/verify_scores.py`, which regrades each
shipped solution with the shipped judge, requires the result to equal the shipped
`reward.json`, recounts the turns from the shipped rollout with
`calibration/audit_trajectory.py`, and re-derives each arm's prompt from
`steps/solve/instruction.md`. It refuses to certify an arm whose prompt has changed since
it ran, which is how the previous round's numbers were kept out of this one.

## What an entry has to say, and why

An entry names the step, when it started and ended, and which of the release's own eight
error tags describes how it was performed, or `"none"`.

The reason is `SPEC.md` section 8 and it does not depend on any score. This task's F1 is
made of two channels and they are not equally hard. A model handed **one still per
recording** and no tools at all places **198 of the 314 steps** correctly by label and
sequence position, which is more than any calibrated agent managed with the whole video
and the ability to seek, where the best was 181. That channel is saturated by a degenerate
baseline. What was left was boundary precision, and boundary precision on this corpus is
bought with decode budget. Requiring the tag makes the score depend on judging the
performance, which no amount of decoding supplies.

The tag vocabulary is the release's, read out of `error_annotations.json` and asserted to
equal what the judge grades. Any tag the annotators named for a step counts, not all of
them: 59 instances carry more than one.

## Why the task is hard, in one table

| out of 314 | Codex | Antigravity | Claude |
|---|---|---|---|
| right label, right sequence position, timing ignored | 181 | 120 | 173 |
| right label, onset inside tolerance | 34 | 24 | 26 |
| **both boundaries and the tag (scored)** | **8** | **3** | **2** |

## Anti-shortcut ablations

Recomputable with `provenance/ablations/run_ablations.py`:

| submission | F1 |
|---|---|
| oracle | 1.0 |
| empty | 0.0 |
| canonical recipe prior, full dish order | 0.0032 |
| random, mean of 400 | 0.0 |
| best spam of any shape | 0.0006 |
| oracle answers filed under the wrong video | 0.0 |

And the three degraded-input runs the ablation gate asks for, real runs of gpt-5.6-sol at
xhigh rather than simulations, each in its own empty working directory, each forced to
answer, each with **zero shell commands**, driven by
`provenance/ablations/run_measured.py` and retained under `provenance/ablations/measured/`:

| degraded input | entries | label+order | F1 |
|---|---|---|---|
| no media at all | 321 | 41 | **0.0** |
| one still per recording, no tools | 310 | **198** | **0.0032** |
| 16 uniform frames per recording, no tools | 271 | 176 | **0.0** |

The stills and the sheets are cut from the image's own baked media, inside a container
built from it, so what the model saw came from the recordings the agent sees.

And, with `--jitter`, where the 0.10 gate sits in seconds. Monte Carlo over 1000 draws,
printed to the precision the estimate has:

| perfect labels, order and tags, Gaussian boundary noise | F1 |
|---|---|
| sigma = 5 s | 0.183 |
| sigma = 8 s | 0.077 |
| sigma = 10 s | 0.051 |

The same curve with the error field answered `"none"` every time, which is the best an
agent that has not judged the performances can do, sits at 0.149 at sigma = 3 s and 0.065
at sigma = 5 s.

## Verification

| check | what it establishes |
|---|---|
| `provenance/verify_key.py` | re-derives the key, error tags included, without importing the builder |
| `provenance/test_oracle_integration.py` | **executes** the shipped `solve.sh` and grades what it wrote: 1.0 |
| `provenance/test_judge_contract.py` | five contract regressions, each with a control that must fail |
| `provenance/take_selection.py` | re-derives the corpus from rules R1-R5 |
| `provenance/audit_error_observability.py` | the checkable half of the error-field observability audit |
| `environment/verify_frozen.py` | a built image against the frozen record of its 22 baked files |
| `calibration/verify_scores.py` | 34 document claims recomputed, with a sentinel and a corrupted-value control |

## Disclosures

**Antigravity's audit is REVIEW REQUIRED, and it is not a path finding.** It wrote a
section headed "Model Knowledge" into its own checkpoint notes, beginning "The videos are
egocentric recordings of recipe execution from CaptainCook with intentional procedural
errors", and listed the six dishes against the recordings they belong to. The prompt rules
that out in as many words. It is reported rather than set aside because the finding cuts
the safe way: with that advantage the arm still placed the fewest steps of the three and
scored 0.0108. Knowing which dishes are in the corpus does not say when a step happened or
whether it was done right, which is what the score is made of.

**The shipped oracle scored 0.0, and nothing caught it.** `make_task_files.py` built each
row from video, id, t_start and t_end, so `solve.sh` carried no error field, while every
check that claimed to grade "the oracle" was grading an error-aware one it had built out
of the key in memory. Those two objects came apart and the checks kept passing on the one
that was never shipped. `test_oracle_integration.py` now executes the shipped script and
grades the `solution.json` that run writes; its control strips the field from a copy and
requires that copy to fall short.

**`memory_mb` went from 8192 to 16384.** This corpus is the heaviest in the family,
`storage_mb` 39936 against 24576 for the next largest, and memory was the one declaration
that had not scaled with it: 13 of the family's 14 tasks write 8192 whatever their corpus.
Under 8192 the Antigravity arm was OOM-killed twice, at 32 and 36 minutes, with `agy`
itself holding 7.28 GiB read out of `/proc`. Headroom can only help an agent, so this
cannot be tuning that makes the task look like it passes; a starved agent scores lower,
and lower is the direction that would. The three scores above were all measured after the
change, not carried across it.

**The tablet CaptainCook4D prompts participants with is often in shot and its step text is
legible.** At native 1080p it is a smear; upscaled 2x it reads back as this key's own label
strings, in canonical order. What it is worth is bounded by the ablation above: the screen
gives labels and order, which is the channel a single still already saturates, and it does
not give boundaries. `provenance/observability/` reads that same screen at native
resolution in two kitchens for the newly scored field and finds the step texts and a
highlight, no error wording, and none of the 352 released step descriptions phrased as an
error instruction. `SPEC.md` open item 6 states the residual in full.

**The error field was audited for observability before it was scored.**
`provenance/observability/` carries one exhibit per tag that can decide a row, shows that
the two categories with no visual definition are never a row's only tag, that no row is
scored against a twin inside its own tolerance, and that the label numbering is
alphabetical rather than recipe order. No row is masked or excluded, and the limitations
are stated rather than left to be found.

Ground truth is derived by `provenance/build_gt.py` from CaptainCook4D's published
annotations rather than authored by us, with assertions against four ways the source could
silently corrupt the key. The tag examples in the prompt are synthetic, written by hand
and asserted to share no four-word run with any of the 8274 descriptions the release ships.


techgenmini:
Hi @Y-Yiqing — thanks for the unusually auditable submission. I rechecked the generated key/oracle, scorer, retained rollouts, and media chain. The released CaptainCook4D annotations are a sound regular
  provenance basis, `allow_internet = false` is correct, and I found no evidence of lookup in the retained runs. I am not requesting a bespoke re-annotation pass.

  I found two task-contract bugs in the inline threads: the prompt exposes three exact GT rows, and equal-start events have a hidden LCS ordering requirement. Please fix those before treating the current
  calibration as final.

  After the contract is frozen:

  - [ ] Run one final current-contract arm each for Codex, Claude, and Antigravity in the built task environment. The retained runs executed task actions on the host and used paths/tools absent from the
  Dockerfile, including `/bin/zsh` and macOS font paths; `ship_rollout.py` then rewrites those roots to `/workspace`. Host-side model transport is fine, but task actions should run through the frozen image with
  no general egress. This is one run per harness, not an n=3 request.
  - [ ] Keep a revision-pinned native raw archive with SHA. Secrets and home-directory prefixes may be redacted, but actual commands, tool outputs, turns, frames, and the host/container boundary should remain
  auditable.
  - [ ] Add a real strong-model, forced-answer `no_media` run. The canonical-recipe controls are deterministic scorer constructions, not a model ablation. Also rerun `single_frame` in genuine no-tool isolation
  and regenerate `frame_dump_no_tools` from the corrected prompt.
  - [ ] Add an automated check that source annotation time maps to the final baked MP4 PTS. Checking only that total duration differs by less than two seconds does not establish boundary alignment.
  - [ ] Freeze the media chain: pin the Hugging Face URL to an immutable revision, and reconcile the checked-in VideoToolbox recipe with the published derivative identifying itself as `libx264`. Either assert
  the expected derivative digest during regeneration or narrow the byte-for-byte reproducibility claim.
  - [ ] Correct the statement that the videos contain no on-screen text. Because the tablet can reveal labels, order, and potentially scroll timing, add a focused tablet-aware measurement or request scoped
  acceptance of that remaining shortcut risk.

  Small cleanup: reconcile the 10800/21600-second manifest mismatch, remove the unused runtime `HF_TOKEN`, and fill/validate the missing A–C width/height fields.

tasks/agentic_vbench_understanding/captaincook-recipe-step-transcripts/provenance/make_task_files.py
```diff
@@ -0,0 +1,245 @@
...
+    vids, inst = d["videos"], d["instances"]
+    n = len(vids)
+    letters = [v["letter"] for v in vids]
+
+    rows = ["| video | path | length | time range |", "|---|---|---|---|"]
+    for v in vids:
+        rows.append(f"| `{v['letter']}` | `/workspace/materials/{v['letter']}.mp4` | "
+                    f"{v['duration_sec']/60:.1f} min | `t = 0` to `t = {v['duration_sec']:.1f}` |")
+    first, last = letters[0], letters[-1]
+    picks = [(first, 0), (first, min(1, len(inst[first]) - 1)), (last, 0)]
```
techgenmini: These schema examples are selected directly from `inst`, so the generated prompt exposes three exact GT events. Regrading only those three rows gives 3 true positives / 0.0189 F1 without watching media.
  Please generate vocabulary-valid synthetic examples that cannot match any selected event, and add an assertion preventing this from recurring.

tasks/agentic_vbench_understanding/captaincook-recipe-step-transcripts/steps/solve/tests/judge.py
```diff
@@ -0,0 +1,630 @@
...
+        onset = float(onset)
+        offset = float(offset)
+    except (TypeError, ValueError):
+        return None
+    if video not in GROUND_TRUTH:
+        return None
+    return {"video": video, "id": label, "t_start": onset, "t_end": offset}
+
+
+def _max_monotonic(preds, gt, matcher):
```
techgenmini, line 568: The prompt permits either order for entries with equal `t_start`, but this LCS fixes the key's arbitrary order. Swapping U's two otherwise exact 455.647-second rows drops a perfect oracle from 1.0 to 0.9968.
  Please make equal-start groups order-insensitive, or define an agent-usable canonical tie rule, and add a regression test.

tasks/agentic_vbench_understanding/captaincook-recipe-step-transcripts/provenance/ablations/measured/single_frame/transcript.jsonl
```diff
@@ -0,0 +1,8 @@
+{"type":"thread.started","thread_id":"01a0492b-8255-77b1-8bda-efcf68bb4ebd"}
+{"type":"turn.started"}
+{"type":"item.completed","item":{"id":"item_0","type":"error","message":"Skill descriptions were shortened to fit the 2% skills context budget. Codex can still see every skill, but some descriptions are shorter. Disable unused skills or plugins to leave more room for the rest."}}
+{"type":"item.completed","item":{"id":"item_1","type":"agent_message","text":"I’ll reconstruct each timeline strictly from the recordings, using the stills only as visual anchors. I’ll identify the dish/action set for each clip, then inspect candidate boundaries closely enough to keep both timestamps within the stated tolerances."}}
+{"type":"item.started","item":{"id":"item_2","type":"command_execution","command":"/bin/zsh -lc \"pwd && rg --files . | sed -n '1,160p'\"","aggregated_output":"","exit_code":null,"status":"in_progress"}}
```
techgenmini: This retained “single-frame, no-tools” run invokes a shell and lists a directory containing both the `single/` assets and the richer 16-frame `dump/` assets. Its zero therefore does not establish the
  advertised isolated condition. Please rerun it attachment-only, with tools disabled and only the single-frame assets accessible.

tasks/agentic_vbench_understanding/captaincook-recipe-step-transcripts/environment/Dockerfile
```diff
@@ -0,0 +1,132 @@
...
+# O = 7_50     Breakfast Burritos       687.55 s     493 MiB
+# P = 7_135    Breakfast Burritos       928.89 s     666 MiB
+# Q = 10_42    Pinwheels                748.81 s     536 MiB
+# R = 10_46    Pinwheels                709.24 s     508 MiB
+# S = 10_48    Pinwheels                649.75 s     466 MiB
+# T = 13_38    Butter Corn Cup          816.22 s     585 MiB
+# U = 13_41    Butter Corn Cup          914.61 s     656 MiB
+# V = 13_44    Butter Corn Cup          766.03 s     549 MiB
+
+ARG BASE=https://huggingface.co/datasets/Maxine668/captaincook-recipe-step-transcripts/resolve/main
```
techgenmini: `resolve/main` is a mutable branch URL rather than an immutable media revision. The per-file hashes make future drift fail closed, but the calibrated image would no longer rebuild. Please pin the dataset
  revision hash and carry it through the generator and provenance text.

tasks/agentic_vbench_understanding/captaincook-recipe-step-transcripts/provenance/data_setup/02_prepare_media.sh
```diff
@@ -0,0 +1,194 @@
...
+
+def transcode(video, src, src_sha, src_bytes):
+    letter = video["letter"]
+    side = work / "work" / f"{letter}.json"
+    dst = work / "out" / f"{letter}.mp4"
+    # Hardware decode and hardware encode. The source is 4K HEVC and the CPU path runs
+    # at about half real time, which is most of the wall clock of the whole job. The
+    # bitrate is set high enough that the encoder is not what limits legibility: 6 Mbit/s
+    # at 1080p is above what the CPU encoder produced at CRF 24 on the same material.
+    subprocess.run(["ffmpeg", "-v", "error", "-y", "-hwaccel", "videotoolbox",
```
techgenmini, line 129: This recipe currently specifies `h264_videotoolbox`, while the published A derivative identifies its encoded stream as `libx264`; the script also records a newly produced digest rather than checking it
  against the committed manifest. Please reconcile the actual recipe and expected digest, and add a focused source-time → final-PTS regression. A one-file demonstration is sufficient.

tasks/agentic_vbench_understanding/captaincook-recipe-step-transcripts/SPEC.md
```diff
@@ -0,0 +1,281 @@
...
+# they did in the wrong order, and 199 of the 314 steps are annotated by the dataset as
+# performed with an error.
+#
+# No recording can be answered from a glance and none dominates the key: the largest
+# single recording holds 6.1 percent of it and the largest five hold 27.1 percent, so
+# concentrating on a handful leaves most of the key on the table.
+
+# 2. Which modalities are REQUIRED?
+#
+# Video only, and the video is load-bearing rather than decorative. The clips carry no
```
techgenmini, line 27: This says there is no on-screen text, but the later tablet disclosure establishes that recipe text is legible after upscaling. Please correct the statement. The canonical-order prior bounds labels/order, but
  it does not by itself bound timing inferred from the tablet's live scroll position.

techgenmini:
> Hi @Y-Yiqing — thanks. I rechecked 86b6a78. The two contract bugs are closed, and the replacement forced-answer ablations independently regrade to 0.0 / 0.0031 / 0.0032. The checked-in Dockerfile now also
  > uses an immutable HF revision.
  >
  > The remaining load-bearing item is the final calibration: calibration/verify_scores.py correctly exits non-zero because all three retained full-task arms answered the pre-fix prompt. Please replace them
  > with one current-prompt/current-image run each for Codex, Claude, and Antigravity through the built task image, preserving a revision-pinned native raw archive with hashes. No n=3 campaign is requested.
  > Please then update the PR body, which still presents the stale rows and old ablation table as final.
  >
  > Three bounded media-chain follow-ups from the earlier checklist remain:
  >
  > - [ ] Add the source-annotation-time → final-baked-MP4 PTS regression.
  > - [ ] Make 03_upload_media.sh carry the returned HF revision instead of regenerating resolve/main.
  > - [ ] Fill and validate the current width/height values for A–C in media_manifest.json.
  >
  > The corrected tablet disclosure is sufficient for me to treat scroll timing as a scoped residual shortcut risk, provided the final current-contract runs remain below the gate and their traces show no
  > successful exploitation. I am not requesting another bespoke detector.

Y-Yiqing:
Thanks. Everything on your list is done, but the answer to the load-bearing item is not
the one either of us expected, and it changed the task. The short version first.

**The three media follow-ups are closed.** The source-annotation-time to baked-PTS
regression is `provenance/check_pts_alignment.py`; `03_upload_media.sh` now resolves the
revision the upload produced and pins that instead of `resolve/main`; A, B and C are
re-encoded, published, and carry width/height. All 22 objects on the hub match
`media_manifest.json` digest for digest, and A/B/C now report `h264_videotoolbox` like
the other nineteen.

**The final calibration is done, and it is a different task than the one you reviewed.**
Re-running Codex under the fixed contract inside the image scored **0.1621** against the
0.10 ceiling. That number is not in any table below; it is why there is something to
report at all. Everything that could have made it an artefact was checked: the judge was
unchanged (all three retained solutions regrade to their shipped values to the digit), the
prompt had become harder rather than easier, the container is 1.2 to 1.6x slower per
operation than the host, the agent's own trace records the network as blocked, there was
one thread and no subagent, and nothing holding the key was reachable inside the
container. The difference was effort: the run on record stopped after 72 turns with five
sixths of its budget unused; this one used 213. The 0.0173 was the unrepresentative
measurement, and the difficulty argument this task rested on did not survive a serious
attempt.

**What we changed, and what we deliberately did not.** An entry now has to say how its
step was performed: `"none"`, or one of the release's own eight error tags. A match needs
the label, both boundaries, and a tag the release annotates for that step (any one of
them; 59 of the 314 instances carry more than one).

The reason does not depend on that 0.1621. Your ablation gate already measured it: a model
given sixteen frames per recording and no tools places 192 of 314 steps correctly by label
and sequence position, and every calibrated agent with the full video and seeking lands
within a few of the same number. That channel is saturated by a degenerate baseline, so
the score was decided by boundary precision alone, and boundary precision here is bought
with decode budget. CaptainCook4D's distinguishing annotation is the one the task was
using only as an eligibility filter. Requiring it makes the scored quantity depend on
judging the performance rather than on locating it.

We did not tighten the tolerance, cut the budget, or re-select the corpus. The
one-recording-per-dish corpus that `take_selection.py` records was built and tested for
exactly this and dropped on the evidence: it triples the vocabulary, but the
canonical-recital saturation moves from 68 percent to 70 percent, so it does not touch the
defect. SPEC.md section 8 carries all of this, including that the trigger was a score.
### The numbers

One run per harness, current prompt, current key, current judge. One agent, one session,
no subagents in any of them.

| harness | model | score | tool calls | entries | in the image |
|---|---|---|---|---|---|
| Codex CLI 0.144.1 | gpt-5.6-sol xhigh | **0.0703** | 135 | 312 | yes |
| Antigravity | gemini-3.6-flash-high | **0.0362** | 343 | 238 | no, see below |
| Claude Code 2.1.251 | claude-opus-4-8 | **0.0152** | 145 | 213 | yes |

Ablations, all real forced-answer runs of gpt-5.6-sol at xhigh, each in its own empty
directory, each with zero shell commands: no media **0.0**, one still per recording
**0.0033**, sixteen frames per recording **0.0034**.

The error field is a requirement rather than a cap, and the three arms show it is not a
formality. Replacing each submission's error field with a constant `"none"`, which is the
key's largest single class at 36.6 percent and the best an agent that has not judged the
performances can do:

| arm | as submitted | error field replaced by a constant guess | of its matches: on correct steps / on diagnosed errors |
|---|---|---|---|
| Codex | 0.0703 | 0.0319 | 9 / 13 |
| Antigravity | 0.0362 | 0.0362 | 9 / 1 |
| Claude | 0.0152 | 0.0152 | 4 / 0 |

Only Codex is diagnosing, and most of its credit comes from doing so. The other two are
indistinguishable from a constant guess on that field. It is worth saying plainly that this
also caps how much comfort the margin gives: a submission that never looks at how anything
was done still reaches 0.149 at sigma = 3 s on the jitter curve in `calibration/scores.md`,
so the field moves the crossing point from about sigma = 7 s to about sigma = 3.5 s rather
than making the task safe.
### The Antigravity arm ran on the host, and we are asking you to accept that

Codex and Claude were re-run inside the built image. Antigravity was not, because we
could not find a way to make it execute anything anywhere else, and we would rather show
you what we checked than assert it.

1. **Its execution-target selector offers only the host.** The control beside the model
   picker lists `Local` and `New Worktree`, and the second is greyed out with "Worktrees
   are available for Git repositories". There is no container, remote or sandbox target.
2. **Settings has no shell or terminal path.** It has a terminal auto-execution policy,
   a file-access policy, an allowed-commands list, and an "Enable Sandbox Mode (Preview)"
   toggle described as "Restricts agent tools to a secure, isolated **local** sandbox".
   None of them relocates where a command runs.
3. **It does not honour `SHELL`.** We launched the app with `SHELL` pointed at a wrapper
   that execs `docker exec -i -w /workspace <container> /bin/sh`, then asked the agent to
   run `cat /etc/os-release; uname -srm; ls /workspace/materials | wc -l`. It answered
   `cat: /etc/os-release: No such file or directory`, `Darwin 24.6.0 arm64`, and
   `ls: /workspace/materials: No such file or directory`. The wrapper itself works: run
   directly it prints `Linux 7.0.12-linuxkit aarch64` and `22`.
4. **Its only non-GUI entry point is a macOS binary.** `~/.gemini/antigravity/bin/agentapi`
   is a one-line shell script that execs
   `/Applications/Antigravity.app/Contents/Resources/bin/language_server agentapi`, an
   arm64 Mach-O. Driving it headlessly would still execute on macOS.

We also tried the substitution your other tasks use. `gemini-cli` is in Harbor's agent
list and the minecraft task calibrates its third arm with it, so we installed it and
attempted the OAuth flow. Google refused: "This client is no longer supported for Gemini
Code Assist for individuals. To continue using Gemini, please migrate to the Antigravity
suite of products." That leaves an API key or Vertex, neither of which we have.

**What the host arm therefore is and is not.** It ran in a clean directory containing only
the 22 recordings and the prompt; every clip's SHA256 matches
`provenance/media_manifest.json`; the prompt is `calibration/make_prompts.py` applied to
that directory and substituting the path back reproduces `steps/solve/instruction.md` byte
for byte. Its trajectory is one session with zero subtrajectories, 343 tool calls, and the
harness's own `executor_metadata` records the model as `gemini-3.6-flash-high` across all
344 planner turns. What it is not is a measurement made in the shipped environment: it used
the host's ffmpeg and the host's Python, and that is exactly the substitution you objected
to. We are not arguing it away.

**The request.** Accept this one arm as a scoped exception, on the grounds that the two
arms that decide the gate are in-image and this one is the weaker of the three claims. If
you would rather not, two alternatives we are happy with: drop the Antigravity row and
report two in-image arms, or tell us how the family runs it, since your own tables say
"Antigravity 1.1.8 via Harbor 0.20.0" and the pinned Harbor here is 0.6.6, whose agent
list has no antigravity entry. If 0.20.0 can drive it into a container we will redo the
arm that way.
### The archive, and one question

The revision-pinned native raw archive is built by `calibration/make_raw_archive.py`. It
redacts two things and nothing else: home-directory prefixes, and credentials, and each
rule ships with a canary that must be caught or the build fails. Frames are kept, and the
count of long base64 strings must be identical in and out, so a future edit that starts
eliding them fails rather than passes quietly. A second sweep with different patterns then
looks for credential shapes the first list does not know, also with canaries; two of those
canaries were added after they caught real blind spots. One further redaction is declared
rather than silent: a harness startup row listing the operator's own installed skills has
its names and prose replaced by a note of their size, and the row and the skill count
survive because what the harness injected is part of the environment under audit.

We have not published it yet: it is 134 MB, 99 percent of which is the 300 frames the
Claude arm looked at. Tell us where you would like it and it goes up with its SHA256s.

**The 10800/21600 mismatch.** We could not find the conflict. `build_timeout_sec = 10800`
is the environment build and `steps.agent.timeout_sec = 21600` is the agent step; they are
different phases, and the sibling tasks pair unequal values the same way. If you meant
something else by it, say which two numbers and we will fix it.

One more thing you should know rather than discover. That agent budget was justified in
`task.toml` as sitting above the longest calibration run, and the longest calibration run
was the 344-minute Claude arm that we now know was not representative. The number was
fitted to a weak observation. We have not moved it, because moving it after seeing a score
is the tuning this family warns about, but the justification written beside it is weaker
than it reads and you should decide whether it stands.

techgenmini:
Hi @Y-Yiqing — thank you for the unusually candid and well-documented update. I rechecked db7a5f2.

  Several earlier items are closed:

  - The PTS regression, immutable Hugging Face revision handling, and A–C media metadata are now in place.
  - The forced-answer degraded-input runs independently regrade to 0.0, 0.0033, and 0.0034, all within the published gate.
  - The current Codex, Claude, and disclosed host Antigravity solutions reproduce the reported 0.0703, 0.0152, and 0.0362.
  - I continue to accept the pinned CaptainCook4D annotations as this task’s provenance basis. I am not requesting another annotation team or a wholesale reannotation.
  - build_timeout_sec = 10800 and the 21600-second agent timeout govern different phases, so there is no mismatch to fix. I also do not recommend shortening the agent budget after seeing the score.

  The pivot to scoring the released error taxonomy is reasonable in principle, but it introduced a new contract that needs one bounded validation cycle before the new runs can be final.

  1. Make the new error contract internally self-verifying.

     Running:

     python3 provenance/verify_key.py --cc4d <pinned-annotations> --task .

     currently fails with the judge's key is not what the annotations say, because the independent verifier rebuilds the old step/timing schema but not the newly scored error annotations. Please update it to
     independently derive the accepted error tags from error_annotations.json, and keep the README claim that this is an independent check true.

     I also reproduced two narrow input-validation inconsistencies:
      - an empty or whitespace-only error value normalizes to "none", although the prompt says a missing/invalid error cannot count;
      - zero-length or reversed time intervals can receive credit when both endpoints happen to fall inside a short row’s tolerances.

     Please reject those cases and add focused regressions. This is scorer hardening, not a request to change the matching metric or lower the reported scores artificially.

     Finally, the new error-tag examples are copied from real annotation descriptions in recordings outside the selected corpus. They are not direct leaks of the 22 scored recordings, but the family rule
     requires task examples to be synthetic. Please replace them with synthetic descriptions that cannot reproduce an annotation row.

  2. Add a compact observability check for the newly scored field.

     The earlier tablet discussion was resolved for the old label/order/timing contract. The new revision now scores whether each step was performed with labels such as Measurement Error, Timing Error, and
     Temperature Error, so this changes the premise enough to require a targeted check.

     Please provide a compact per-tag audit, concentrating on the ambiguous categories and rows, showing what in the supplied video lets the agent distinguish the annotated error from the intended action.
     Please also check the opposite shortcut: whether an induced-error script or readable tablet content directly reveals the label without requiring the agent to judge the performance.

     This does not need to become a new 314-row manual annotation project. Representative evidence per category plus row-level handling of cases that the audit flags is sufficient. For any retained row whose
     tag cannot be supported from agent-visible evidence, use masking, exclusion, or another explicit treatment rather than forcing the hidden label.

  3. Freeze the result before spending on the final campaign.

     Please resolve the key, scorer, prompt, and observability items first. After that freeze, retain one final current-contract run each for Codex, Claude Code, and native Antigravity, plus the required
     degraded-input runs. No n=3 campaign is requested.

     The current host Antigravity row is honestly documented, but it does not close the earlier in-image requirement. The repository already has merged Harbor 0.20 / native agy examples for the Dota 2 and
     IndustReal tasks. Please try that installed-agent path inside the task image before requesting an exception. If it fails because of a concrete authentication or adapter limitation, preserve that evidence
     and we can make a narrowly scoped maintainer decision rather than treating the GUI experiments as the final word.

     Because another campaign is required anyway, freeze the base image before it and record the resulting task-image digest with the final run metadata. This is part of the final campaign, not a separate build
     round.

  On the 134 MB archive question: please do not upload the current bundle yet, since the work above may make those runs non-final. Once the contract is frozen, either a revision-pinned Hugging Face dataset
  location or an immutable GitHub release asset is acceptable. Record each archive’s URL, revision, byte size, and whole-file SHA-256 in the repository.

  Credentials and home-directory prefixes may be redacted. For the injected local skill inventory, the public copy may omit private names and prose if that transformation is precisely declared, the original
  field is bound by a digest, and the unmodified source remains available for maintainer-only audit. Commands, outputs, turns, frames, failures, and the host/container boundary must remain auditable.

  After the final evidence is settled, please update the PR description and task documentation to the new error-aware contract and final scores. I am not reopening media provenance, requesting corpus
  reselection, tightening the score threshold, or asking for additional annotators.

  For transparency, I ran the independent key/scorer/calibration checks and inspected the pinned media chain, but I could not perform a fresh full Docker build because the local Docker daemon is unavailable.

techgenmini:
> Small Docker follow-up on the same head, db7a5f2, and a correction to my last transparency note: Docker is now available, and I completed a fresh arm64 build and runtime check.
  >
  > The ordinary default build succeeded without overrides. All 22 revision-pinned inputs matched their declared SHA-256 values and passed the post-remux sanitization checks. The setup handoff passed under
  > --network none, exactly 22 non-empty videos were presented in /workspace/materials, their hashes matched the corresponding /baked files, and all 22 agent-visible videos completed a strict full decode. This
  > closes the build and media-runtime uncertainty from my previous review.
  >
  > The end-to-end check did uncover one additional merge blocker in the generated oracle path. steps/solve/solution/solve.sh still emits the old 314-row schema without the newly required error field. Running
  > that shipped oracle through the current shipped verifier produces reward 0.0: all 314 predictions align on label, order, and timing, but none can become a true positive without error.
  >
  > The existing contract regression does not catch this drift because it constructs a separate error-aware oracle in memory. The source of the generated artifact has the same omission: provenance/
  > make_task_files.py currently builds each oracle row from only video, id, t_start, and t_end.
  >
  > Please fix the oracle generation path so each generated row carries a deterministic accepted error value, refresh steps/solve/solution/solve.sh from that generator, and add an integration regression that
  > executes the shipped oracle and asserts reward 1.0. This is a generated-artifact/integration correction; it is not a request to regenerate the ground truth or media.
  >
  > One media-identity clarification: media_manifest.json authenticates the hosted derivative inputs, while the agent-visible /baked files are post-remux outputs and therefore have different whole-file hashes.
  > I am not requesting another encoding or publication step. This simply reinforces the existing request to freeze the base and record the final task-image digest after the contract is settled.
  >
  > This supplement does not otherwise expand or change the bounded checklist in my previous comment.

Y-Yiqing:
Thank you. Everything in the bounded checklist is done, and the supplement's oracle defect
is fixed with the integration regression you asked for. Head is `2c5fba9`.

## 1. The error contract verifies itself

`verify_key.py` now rebuilds the error field too. It re-reads `error_annotations.json`,
joins on `(step_id, start_time)`, and asserts the join is total, so a step that finds no
error row fails instead of defaulting to `"none"` and reading as correctly performed. It
also requires the judge's taxonomy and the prompt's offered values to equal what the
release annotates. Four injected faults were confirmed to fail it. The README claim that
this is an independent check stays true: it still does not import the builder.

Both input-validation findings were real and are closed. An empty or whitespace-only
`error` no longer normalizes to `"none"`; on otherwise perfect timing that was worth
0.3662, and the prompt already said a missing value cannot count. A zero-length or
reversed span is rejected in `_norm`, so it costs precision like any other malformed
entry; six of the key's 314 rows are short enough that one instant satisfied both
boundaries. Neither moved the arms, because none of them tripped either.

The tag examples are now synthetic, written by hand in a dish CaptainCook4D does not
contain. The guarantee is asserted rather than intended: no example may contain, be
contained by, or share a four-word run with any of the 8274 descriptions the release
ships, checked against every recording rather than only the 22.

## Supplement: the oracle

You are right, and the diagnosis is exactly right. `make_task_files.py` built each row from
`video`, `id`, `t_start` and `t_end`, so `solve.sh` carried no error field, and every check
that claimed to grade "the oracle" was grading an error-aware one it had built out of the
key in memory. Those two objects came apart and the checks kept passing on the one that
was never shipped.

The generator now emits a deterministic accepted tag and `solve.sh` is regenerated from it.
`provenance/test_oracle_integration.py` **executes** the shipped script, reads back the
`solution.json` that run wrote, and requires 1.0. Reading the SEQUENCE literal back out of
the file would have repeated the same mistake one level down, so it runs the script, either
natively where `/workspace` is writable or in a container, and a run that cannot happen
fails rather than passes. Its control strips the error field from a copy of the same script
and requires that copy to fall short; it scores 0.0 with all 314 rows still right on label,
order and timing, which is your finding restated.

## 2. Observability of the newly scored field

`provenance/observability/` has the audit, with one exhibit per tag and a script that
recomputes the structural half. Two results carry as much as the exhibits:

**`Other` and `Missing Step` are never a row's only tag.** Every row carrying either also
carries a tag with a visual definition, and any annotated tag scores, so no row turns on a
catch-all with no visual signature. **No row is scored against a twin it cannot be told
apart from:** two row pairs sit inside each other's tolerance on both boundaries, and
neither disagrees on the tag. The one row where Temperature Error stands alone is the one
with a thermometer reading `79.4` on screen while the step says to check boiled water.

The finding that decided the rest: the evidence is in the pixels but not in a downscaled
whole frame. A microwave keyed to `3:00` for a four-minute step is legible at native
resolution and unreadable when the frame is scaled to fit a vision model. The agent has
ffmpeg and can crop; knowing where and when to look is the work.

Reverse direction: only `instruction.md` names the tags, uniformly and tied to no row. The
tablet, read at native resolution in two kitchens, carries the step texts and a highlight
and no error wording, and none of the 352 released step descriptions is phrased as an
error instruction. During the audit I misread a blurred tablet line as an induced-error
instruction and checked it against the corpus before reporting it; the line says something
else. That is in the write-up.

**No row is masked or excluded.** Every load-bearing category has exhibited evidence, and
the limitations are stated rather than left to be found.

## 3. The frozen campaign

Base pinned by digest, checked layer for layer against the image the calibration ran on.
Task image `sha256:de299ec5`, recorded in `environment/frozen.json` with the SHA-256 of
each of the 22 baked files; two builds of the Dockerfile produced byte-identical `/baked`.

| harness | model | effort | score | tool calls | in the image | audit |
|---|---|---|---|---|---|---|
| Codex, codex-cli 0.144.1 | gpt-5.6-sol | xhigh | 0.0274 | 203 | yes | clean |
| Antigravity, Harbor 0.21.0 / CLI 1.1.22 | gemini-3.6-flash | high | 0.0108 | 201 | yes | review required |
| Claude Code, 2.1.251 headless | claude-opus-4-8 | default | 0.0073 | 143 | yes | clean |

Degraded input, all forced to answer, all with zero shell calls: no media 0.0, one still
per recording 0.0032, sixteen frames 0.0.

**The Antigravity exception request is withdrawn.** It was about the desktop application,
which is a different artifact from the CLI Harbor drives, so it answered the wrong
question. The in-image path works and the arm above ran through it.

Getting there needed two flags Harbor does not pass, and the reason is worth recording
because it will affect anyone calibrating this family from now on. No released Harbor does
both halves. Up to `0.21.1.dev202608192356` the adapter seeds the pre-provisioned OAuth
token, which is the only headless credential `agy` accepts, but never passes `--effort`,
and `agy` 1.1.22 refuses to start without it. The next release adds `--effort` and removes
the OAuth seeding in the same commit. Every release in the range was checked. Both flags
are added by a four-line subclass through the `module:ClassName` agent path `--agent`
documents, not by patching the library. The second flag came from a probe: `agy
--print-timeout` defaults to five minutes and knows nothing about the budget Harbor grants.

Also worth flagging for the family: Harbor's adapter copies `agy`'s transcript from the
path 1.1.8 used, and 1.1.22 writes it under `.gemini/antigravity-cli/brain/<session>/`.
One of our runs finished with a score and no auditable record before we found that. The
merged tasks used 1.1.5, 1.1.8 and 1.1.12, so nobody has hit it yet; the installer always
pulls latest.

`task.toml` is unchanged on network policy. `allow_internet` stays false and Harbor's
phase overrides carry the difference on the command line, so both host lists are in the
record. The install hosts came from a traced run of the vendor's installer; every
agent-phase host was added only after a run named it in an error, starting from empty.

## Two changes you did not ask for, both disclosed

**`reward.json` was the wrong shape and would have broken any Harbor run of this task.** It
was `{"reward": ..., "details": {...}}`, and Harbor parses that file as
`dict[str, float | int]`, so the nested object became a second reward whose value is a dict
and the trial died in a `ValidationError` after the agent had already run. It is now a flat
mapping with the breakdown in `details.json` beside it, which is the shape the merged Dota 2
task already uses.

**`memory_mb` went from 8192 to 16384.** This is the one change that touches the shipped
contract, so here is the whole reasoning. Under 8192 the Antigravity arm was OOM-killed
twice, at 32 and 36 minutes, with `agy` itself holding 7.28 GiB read out of `/proc`. Rather
than re-run until one survived, we looked at the declaration: this corpus is the heaviest
in the family, `storage_mb` 39936 against 24576 for the next largest, and memory is the one
declaration that had not scaled with it, since 13 of the family's 14 tasks write 8192
whatever their corpus. Headroom can only help an agent, so raising it cannot be tuning that
makes the task look like it passes; a starved agent scores lower, and lower is the
direction that would. The number that needed defending was the old one.

The order matters and I would rather state it than have you find it: the declaration
changed partway through this round, after two failed Antigravity attempts. All three arms
above were measured after the change, so they are comparable to each other, and none of the
earlier partial numbers is reported anywhere.

## The Antigravity audit finding

It is REVIEW REQUIRED, and it is not a path finding like the last one. The arm wrote a
section headed "Model Knowledge" into its own checkpoint notes, beginning "The videos are
egocentric recordings of recipe execution from CaptainCook with intentional procedural
errors", and listed the six dishes against the recordings they belong to. The prompt rules
that out in as many words.

We report it beside the row rather than setting the arm aside, because the finding cuts the
safe way: with that advantage the arm still placed the fewest steps of the three, 120
against 181 and 173, and scored 0.0108. Knowing which dishes are in the corpus does not say
when a step happened or whether it was done right, which is what the score is made of. If
you would rather the row were dropped or re-run, say so and we will.

## The archive

Published as an immutable release asset, since the contract is now frozen:

    https://github.com/Y-Yiqing/agentic-vbench/releases/tag/raw-calibration-2026-09-01
    104,304,992 bytes, sha256 82110af9d8c609509fbb12f791af6a7db1171b081949e89c8d012c6ba8a56526

`calibration/raw_archive.json` records the URL, revision, byte size, whole-file digest and
a per-arm breakdown. The asset was downloaded from the published URL and re-hashed, so the
recorded digest is the digest of what you will receive rather than of what was uploaded.

Redaction covers home-directory prefixes and credential shapes and nothing else. Every rule
was proven to fire on a canary carrying its target before the run, because a rule that
matches nothing would otherwise report a clean archive, and an independent sweep over the
finished archive caught 6 of 6. As it happens nothing in these three transcripts required
redacting, so the declared-transformation and digest-binding conditions you set for the
skill inventory do not arise: there is no omitted field. Commands, outputs, turns, failures,
the host and container boundary and all 396 image blobs are carried through in full.

## Honest note on the numbers

Every arm scored lower than last round, 0.0274 against 0.0703, 0.0108 against 0.0362,
0.0073 against 0.0152. The contract got stricter, which is the expected direction, but the
run-to-run spread on these arms is wide enough that the size of the drop should not be read
as a measurement of how much the error field costs. The per-arm decomposition in
`scores.md` is the number that speaks to that: replacing Codex's error field with the
constant `"none"` costs it two thirds of its score, and the other two arms are
indistinguishable from that constant.

PR description and task documentation are updated to the error-aware contract and these
scores.

techgenmini:
> Hi @Y-Yiqing — thank you for the thorough response. I completed the re-review of live head 2c5fba9. The substantive bounded checklist is now closed. I am not requesting another calibration campaign or
  > reopening the accepted media, provenance, corpus, timeout, or annotation decisions.
  >
  > I independently verified:
  >
  > - verify_key.py now reconstructs the error annotations from the pinned release, and the generated key, prompt, judge, and oracle reproduce correctly.
  > - The blank-error and invalid-interval regressions pass. The shipped solve.sh executes successfully with reward 1.0, while the error-stripped integration control scores 0.0.
  > - The retained solutions regrade to Codex 0.0274, Antigravity 0.0108, and Claude 0.0073; the forced-answer degraded-input results and all 34 calibration claims also verify.
  > - The published raw archive downloads as exactly 104,304,992 bytes with SHA-256 82110af9d8c609509fbb12f791af6a7db1171b081949e89c8d012c6ba8a56526. Its per-arm hashes, line counts, and image-blob counts match
  >   the repository record, and the archive’s positive controls and independent credential sweep pass.
  >
  > I also accept the retained Antigravity result. Its disclosed model knowledge identified the source and dishes, but I found no network, grounding, key, or grader access. With that advantage it still placed
  > the fewest steps and produced the lowest score, so the finding is conservative for the difficulty claim. I am not requesting a rerun solely for this.
  >
  > On the new field’s observability, the audit closes the category-level feasibility and reverse-shortcut questions: Other and Missing Step are never sole tags, the tolerance-overlap pairs do not disagree on
  > tags, the sole Temperature row is visibly supported, native-resolution crops demonstrate relevant evidence, and the tablet/task files do not expose row-specific error wording.
  >
  > The audit nevertheless records a real longer-term cleanup: some sole-Measurement rows require distinctions such as 1/8 tsp versus 1/4 tsp, although measuring graduations were not legible, and some source
  > labels admit a reasonable alternative under the agent-facing examples. For example, J/5_35 at 684.643–700.783 describes “a big amount instead of a small amount” but accepts only Technique Error, while
  > B/3_46 at 208.442–229.890 treats using a hand instead of a utensil solely as Preparation Error.
  >
  > I do not think this needs to block the PR. We have identified possible ambiguity, but not demonstrated that a specific retained row is wholly unobservable. More importantly, even granting every currently
  > boundary-correct match regardless of its error value gives upper bounds of approximately 0.0377, 0.0251, and 0.0145 for Codex, Antigravity, and Claude—still safely below the gate.
  >
  > Please open a linked follow-up issue with this bounded scope:
  >
  > - inspect the 20 sole-Measurement rows and only the directly implicated Technique/Preparation overlaps;
  > - record row-level visible evidence where it resolves the label;
  > - accept supported alternative tags where the taxonomy remains ambiguous;
  > - neutralize, mask, or exclude the error field where the required distinction is not agent-visible;
  > - add corresponding verifier/scorer regressions and regrade the retained solutions.
  >
  > This should not expand into a 314-row reannotation project. Whether new model runs are required can be decided from the actual semantic changes after the follow-up regrade.
  >
  > Three small evidence corrections remain on this PR:
  >
  > - instruction.md says the examples were taken from elsewhere in the annotation set, and SPEC.md says they came from recordings outside this corpus. They are now synthetic; please make those descriptions
  >   accurate.
  >
  > - claude-manifest.json records Claude Code 2.1.129, while the native trajectory and pinned installer record 2.1.251. run_in_image.py currently captures the host CLI version before entering the container;
  >   please record the in-container version and correct the manifest.
  >
  > - The Antigravity auditor reports 149 absolute paths outside /workspace, all under its own .gemini/antigravity-cli brain/scratch state. They do not appear to expose task or host data, but please classify
  >   them as harness-owned runtime state or disclose them alongside the model-knowledge finding.
  >
  > None of those three corrections requires another model run: they correct provenance wording and retained-run metadata without changing the vocabulary, examples, key, scorer, or submitted answers.
  >
  > Once the follow-up is linked and these small evidence corrections are verified, I expect the final recheck to be short and would recommend approval.

Y-Yiqing:
Thank you. All three evidence corrections are in at `15e2984`, and the follow-up is
#118 with the scope you set.

**The examples.** The prompt and `SPEC.md` both said the wording came from the annotation
set. They are synthetic and have been since the previous round; only the sentence
describing them was stale. Both now say what they are, and SPEC records that the older
description was true of an earlier version and is not true of this one. The three retained
ablation prompts still carry the older sentence on purpose: they are the record of the
question those runs were actually asked, and rewriting them would make the record false.
Regenerating one today differs from the retained copy in that sentence and nothing else.

**The Claude version.** You are right about the cause. `run_in_image.py` read
`claude --version` on the host, so the manifest named this machine's 2.1.129 for a run that
used the pinned 2.1.251 its own trajectory carries as `claude_code_version`. It now reads
the version from inside the container after install and records where it read it. Both
retained manifests are corrected; `scores.md` and the PR description already said 2.1.251,
so they now agree rather than disagree.

**The 149 paths.** All 149 are under `/root/.gemini/antigravity-cli/`: the CLI's brain
directory, the scratch directory where it wrote the frame-extraction scripts and contact
sheets it built for itself, and its logs. `HARNESS_SCRATCH` now names those prefixes, so
the auditor lists them as the agent harness's own runtime state instead of counting them
against the run, and the residual is zero for all three arms. The prefixes are named rather
than matched loosely, because widening that list is the one edit to this auditor that makes
it report less. Antigravity stays REVIEW REQUIRED on the model-knowledge finding alone.

**One thing that fell out of the first correction, worth stating.** Changing the prompt's
sentence moved its digest, and `verify_scores.py` did what it is for: it marked all three
arms stale and refused to certify either gate. Rather than weaken that check, it now
carries `PROMPT_EDITS_SINCE`, an exact before-and-after list of edits that changed how the
prompt describes itself rather than what it asks. An arm's own prompt is reconstructed by
undoing those edits and its recorded digest must still match, so the exemption is by exact
string and cannot cover an edit nobody wrote down. A control confirms the check still
bites: dropping one label from the vocabulary marks all three arms stale again. If you
would rather the arms simply be re-run than have that mechanism exist, say so and we will
re-run them; it is one evening.

I also recomputed your upper bound independently. Granting every boundary-correct match
regardless of its error value gives 0.0377, 0.0251 and 0.0145, agreeing with your figures
to the digit.

Current state at `15e2984`: three arms 0.0274 / 0.0108 / 0.0073 at 203 / 201 / 143 turns,
all in the frozen image; degraded input 0.0 / 0.0032 / 0.0; the shipped oracle executes to
1.0 and its error-stripped control to 0.0; 34 document claims recomputed; the archive
downloads to the recorded digest.

techgenmini:
> Hi @Y-Yiqing — I rechecked live head 15e2984. All three evidence corrections are closed, and linked issue #118 matches the bounded follow-up scope from the prior review. The prompt exception is limited to
  > the exact documentation-only sentence: it reconstructs each retained run’s recorded prompt digest, while an unrelated vocabulary edit still marks every arm stale. The focused verification suite passes, and
  > no model rerun is needed. From my side, this is ready for approval and merge.

PR #117, opened 2026-08-29
Add OpenTTGames full-match rally event-chain reconstruction task

DDMeer:
## Summary

Adds a full-match OpenTTGames table-tennis understanding task that requires reconstructing rally event chains from the provided match video.

The task asks the agent to identify, for each benchmark rally:

- serve timestamp, player, hand, and stroke type
- subsequent visible racket-ball contacts with player, hand, stroke type, and timestamp
- rally-ending label and timestamp

The task uses the official OpenTTGames `game_2.mp4` video and a commit-pinned copy of the published `game_2.json` annotations as the source ground truth, with bounded video auditing for source inconsistencies and observability gaps.

## Ground truth

Current frozen benchmark:

- **92 benchmark rallies**
- **387 benchmark strokes**
- **399 published source stroke annotations**
- **6 source terminal exceptions corrected by bounded video audit** (rally IDs 13, 16, 18, 24, 55, 73)
- **1 documented video-gap exclusion** (rally 8; 12 strokes)
- **0 silently excluded serve windows**
- **0 strokes added by hand** — every benchmark stroke remains source-derived

The six source windows that lacked a usable terminal annotation were reviewed directly against bounded video clips and given explicit terminal labels while preserving the published stroke sequence. The compact audit is in `calibration/source-exception-audit.md`.

A separate issue was found in rally 8: one published serve window spans two points. The first point is retained and terminates at a visually verified left-side net event at frame 12325 (`102.708 s`, the frame at which the net arrests the ball's forward motion).

The following point begins with a right-side serve, but the exact racket-ball serve contact is not present in the source video frames. Because the task contract requires observable serve timing and required serve attributes rather than inferred values, that second point is documented and excluded from benchmark scoring rather than silently guessed.

The benchmark contract requires reporting **visibly observable racket-ball contacts only**. Off-frame contacts are not inferred from ball trajectory alone.

## Provenance

Annotations are pinned to:

- repository: `moamal01/table_tennis_data`
- commit: `36471a76b969a0340df59258a813bf8214e68e7c`
- path: `data/raw/game_data/train/game_2.json`
- annotation SHA256: `7466f1f8c46316406ae224a17491354eac89f9cc2de858633b6f893573db4fe7`

Media:

- official OpenTTGames `game_2.mp4`
- SHA256: `330ac07730bae6d899dbbbd00ad43500c583e6af6ea6dd261565bc77811eba66`

The ground-truth tier is recorded as **machine truth from published structured annotations, with bounded video-audited terminal exceptions** — the transformed key is not presented as purely machine-derived.

The source annotations and all benchmark-specific corrections and exclusions are preserved under `calibration/` for auditability and deterministic regeneration.

## Verifier / regression validation

The verifier is hardened against symlink and non-regular-file submissions (`O_NOFOLLOW | O_NONBLOCK`, `fstat`, size cap) and scores the required stroke attributes jointly, so no single required dimension can be wrong and still earn credit.

Current regression results — all 13 assertions pass:

- oracle: `1.000000`
- empty: `0.000000`
- malformed: `0.000000`
- symlink: `0.000000`
- shifted timestamps: `0.000000`
- wrong label / player / hand / stroke: `0.000000`
- wrong ending label / ending time: `0.000000`
- missing hand: `0.000000`
- sparse one-stroke submission: `0.147559`

`git diff --check`, task structure checks, deterministic GT regeneration, Python compilation, and all verifier regression assertions pass.

## Calibration status

Complete. Ground truth and the six-file task contract were frozen at `6aa6786`, and those six files remain byte-identical to that base. Calibration and ablation-specific execution details are disclosed below.

That freeze commit was `1b7cbd5` when the runs were made; rebasing onto `upstream/main` to clear a `.gitignore` conflict rewrote the commit objects, so the same content now lives at `6aa6786`. Run-time metadata still records the pre-rebase SHA, because editing a recorded artifact to match a later history rewrite would falsify it.

`scripts/understanding/check_task.py` passes all eight checks it can run here: structure, oracle `1.0`, null baseline `0.0`, strong agent `0.041168 < 0.1`, `123 > 50` tool-call turns, and the three required ablations at `0.0 <= 0.15`.

### Agents

| harness | version | model | reasoning | reward | tool-call turns | wall time | trajectory |
|---|---|---|---|---:|---:|---:|---|
| Claude Code CLI | 2.1.251 | `claude-opus-4-8` | default | **0.041168** | **123** | 101.0 min | `rollouts/claude_claude-opus-4-8.jsonl` + `.seg2.jsonl` |
| Codex CLI | 0.147.0 | `gpt-5.6-sol` | high | **0.000078** | **150** | 35.5 min | `rollouts/codex_gpt-5.6-sol.jsonl` |
| Antigravity CLI | 1.1.22 | `gemini-3.1-pro-high` | high | **0.000000** | **164** | 40.6 min | `rollouts/antigravity_gemini-3.1-pro-high.native.jsonl` |

Wall time is the span from a run's pre-`net_guard` timestamp to its post-`net_guard` timestamp, so it includes the CLI install phase and every value is computable from the committed `*.netguard.log` files alone — Codex `21:51:43Z → 22:27:11Z`, Antigravity `09:32:42Z → 10:13:21Z`, Claude `69.3 + 31.7` across its two segments. Per-field diagnostics, the per-harness turn-counting rules, and the isolation and lookup-audit tables are in `calibration/scores.md`.

The strongest run reaches rally discovery F1 `0.854` and stroke timing F1 `0.743` but only `0.260` on joint stroke semantics and `0.250` on joint rally endings. Finding the rallies is tractable; reconstructing each contact's player/hand/stroke and the terminal event is what the task actually measures, and reward is the product of the four terms.

The Claude row is one session executed in two segments, split by a subscription-window interruption and resumed with `claude --continue`. Both segments carry `session_id` `40ddfba8…` and both raw transcripts are committed, because segment 2's 37 turns alone would not clear the `>50` gate — the reported 123 is `86 + 37`.

### Ablations

| ablation | harness | reward | tool calls | submitted (rallies / strokes) | matched (of 92 / 387) |
|---|---|---:|---:|---:|---:|
| `single_frame` | Codex | **0.000000** | 4 | — | — |
| `no_media` | Codex | **0.000000** | 3 | — | — |
| `no_media`, forced answer | Codex | **0.000000** | 6 | 96 / 467 | 19 / 54 |
| `frame_dump_no_tools` | Claude Code | **0.000000** | **0** | 43 / 158 | 3 / 5 |

All four clear the `0.15` gate, but they are not interchangeable. `single_frame` and plain `no_media` are abstention controls: the agent declined to answer, which bounds completability without the media. The other two are the non-abstaining evidence.

**Forced `no_media` bounds recall.** Told to guess anyway, the agent produced a full-match-shaped answer — 96 rallies and 467 strokes spanning 63.4 s to 1388.5 s — and matched 19 of 92 rallies and 54 of 387 strokes. Guessing the shape of a match is easy; landing serve contacts inside 1.0 s and stroke contacts inside 0.35 s is not.

**`frame_dump_no_tools` is the literal condition.** The complete selected sample — all 1435 frames at 1 fps, none omitted — is pre-arranged into 30 7×7 contact sheets at 1568×882 and handed to the model as image inputs on a single request with every tool disallowed. No shell, no file access, no inspection of any kind; `tool_use` count in the trajectory is 0. It submitted 43 rallies and matched 3.

Sheet geometry was chosen by measurement. Image cost tracks pixel area rather than tile count, so 7×7 and 5×5 cost the same per sheet and 7×7 simply needs fewer of them, leaving real output room inside the context window.

The sheets are reproducible, not merely hashed. `ablations/ablation_frame-dump-notools_sheets.sha256` pins which 30 the scored request saw, and `runpack/build_frame_dump_sheets.sh` regenerates them from the pinned source media: it verifies the media digest, extracts 1 fps frames, tiles 7×7 at 224×126, encodes JPEG at `-q:v 5` through an asserted ffmpeg `7.1.5-0+deb13u1`, and finishes with `sha256sum -c` against that manifest. A clean run reproduces **all 30 committed digests**. No JPEG binaries are committed. The Docker image ID is recorded but not pinned, because image IDs are not reproducible across independent builds of the same Dockerfile and pinning one would reject a reviewer who built the image themselves.

On legibility: at 224×126 per tile the players, stances, racket arms, table and net all read clearly, and the model's probe reply described the ball as a resolvable dot, but direct inspection of a mid-rally tile does not corroborate that — the ball is a few pixels and is not reliably identifiable by eye. That question is left unsettled. What the run does show is that the presentation was usable enough to work from: it submitted 158 stroke times, 73% of them non-integer, so it interpolated between sampled instants rather than echoing them, and still matched only 3 rallies and 5 strokes.

The earlier frame-dump run — same frames, but on disk, with the agent spending 44 shell/Python calls inspecting them — was a frames-plus-tools condition and is no longer reported. It submitted 89 rallies and matched 10 rallies and 7 strokes; a submitted count is not recall. It is retained locally under `ablations/superseded/` as history.

### Ablation-only rule overrides, and why they differ

The two non-abstaining diagnostics only work if the agent will predict without having observed. The shared rules forbid exactly that — rule 4 is "work from the video" — and the agents obeyed: the first forced `no_media` attempt returned one placeholder rally, the second an empty list citing "binding workspace rules", and the first `frame_dump_no_tools` attempt did the same. They were right to.

Each needed an ablation-only override, and the mechanisms are different, so they are documented separately rather than as one claim:

- **Forced `no_media`** appends the override to the **container copy** of the rules and records both digests in its metadata — repository `779eec27…`, container `fe217b24…`.
- **`frame_dump_no_tools`** edits no rules file at all. It runs from `/tmp`, which has no `CLAUDE.md`, so the shared rules never enter that context; the override lives only in the prompt text, and its metadata says so and records no container digest.

The repository rules file is unmodified at `779eec27…` in both cases, every scored calibration run used it as-is, and the frozen task contract is untouched.

### Isolation

Scored containers use a default-DROP egress allowlist rather than a blanket internet switch, since `allow_internet` is deprecated in Harbor 0.22.0 and the frozen `task.toml` was not modified for calibration convenience. The dataset host, GitHub, the raw-content CDN and search engines are unreachable; each harness's model transport stays open, because that is inference rather than lookup. `net_guard.sh` re-proves both halves from inside the namespace before and after every scored run — all passed — and the per-run logs ship as evidence for all three harnesses.

Trajectory scans are clean on dataset, annotation, answer-key and search markers for all three runs and all four ablations, with base64 image blocks stripped first so random base64 cannot match a short marker by chance.

Image identity is stated at the strength the evidence supports. The local image is `sha256:db657366…`, created 2026-08-29T23:45:38Z, with no registry digest because it is built locally from the frozen Dockerfile. Only one retained run records that ID directly in its own metadata — the `frame_dump_no_tools` ablation, produced after the review asked for it. The four harness runs recorded the image *name* only, and the forced `no_media` ablation recorded no image line. For those, the claim is an inference rather than a run-time record: the image was built before every reported run with no rebuild in between, which is consistent with all of them having used it, but it was not captured at the time. Reruns from here record the ID directly.

The Antigravity row used the pre-finalization shared rules; the six frozen task files were identical and the only difference is one sentence the finalized file drops. That difference was reviewed and accepted at head `afe9997` without a strict-parity rerun.


techgenmini:
Hi @DDMeer — thanks, this has a strong foundation. I regenerated the 86-rally / 384-stroke key from the pinned upstream annotation, confirmed oracle `1.0` and null `0.0`, verified the annotation SHA against
  the cited commit, and confirmed that the official media endpoint is live and matches the declared size.

  Before merge, I think these bounded items remain:

  - [ ] Harden solution loading. Under the Harbor layout, an agent can leave `/workspace/output/solution.json` as a symlink to `/tests/reference.json`; `judge.py:435-438` follows it during verification and
  awards `1.0`. I reproduced this locally. Please open with `O_NOFOLLOW | O_NONBLOCK`, verify a size-capped regular file with `fstat`, treat invalid inputs as zero, and add a symlink regression test.

  - [ ] Make every required dimension load-bearing. An otherwise-oracle answer with every stroke `player` wrong currently scores `0.666667`, while one with every ending label wrong scores `0.5`. Please use
  full-tuple/joint hits or another zero-preserving aggregation for stroke semantics and ending label+time, with regressions for an entirely wrong/missing dimension.

  - [ ] Reconcile the six silently excluded serve windows: IDs 13, 16, 18, 24, 55, and 73 (`195.158`, `218.792`, `236.108`, `331.967`, `771.217`, and `1057.983` s). They contain 15 annotated strokes and are
  dropped only because the source lacks terminal labels. I spot-checked the video; these include real live-play exchanges, so source-label presence is not an agent-observable definition of “valid.” Please
  provide a compact video audit and either complete/cross-check the reportable endings or define and justify an observable exclusion rule for each window. I’m comfortable accepting the published, commit-pinned
  annotation as this task’s scoped provenance basis; this is not a request to relabel all 384 strokes. Please record that exception in SPEC instead of the non-template `dataset-provided` tier.

  - [ ] Freeze the agent contract before calibration. Define `forehand`/`backhand`, the eight stroke techniques, the ending prefix convention, each ending category, and its timestamp anchor. The upstream
  dataset documentation already provides the ending rules. Also replace `instruction.md:77-100`: it copies the first GT rally’s exact timestamps and labels. A synthetic example outside all scored tolerances is
  enough, and please close the open JSON code fence.

  - [ ] Once those changes are frozen, complete the SPEC difficulty/anti-shortcut sections and add `calibration/scores.md`: one current-image, `allow_internet=false` run each for Codex, Claude Code, and
  Antigravity, with raw native trajectories, harness/model versions, scores, and tool-call turns. Measure single-frame, forced-answer no-media, and all-frames/no-tools. No n=3 campaign or audio ablations are
  requested.

DDMeer:
@techgenmini Thanks for the detailed review — I’ve addressed the bounded items you flagged and pushed the updated task/calibration evidence.

The verifier hardening, load-bearing scoring changes, source-window audit, and calibration/ablation updates are now included. The six frozen task files remain byte-identical to the frozen baseline.

I also documented the calibration methodology and caveats in calibration/scores.md. Please let me know if there’s anything else you’d like tightened before merge.

techgenmini:
> Hi @DDMeer — thank you for the thorough follow-up. I re-reviewed the current head, afe9997b87cdedc67c64876a941389da43a01403. Most of the original checklist is now closed.
  >
  > I verified the pinned upstream annotations and the regenerated 92-rally / 387-stroke key, reviewed the previously excluded terminal windows, and confirmed the instruction/scorer fixes. I also completed a
  > fresh ordinary Docker build without source overrides: the pinned media hash passed, offline setup succeeded, the agent-visible copy matched the baked source, the full media decoded cleanly, and the verifier
  > tests passed.
  >
  > Two bounded calibration items remain before I would call this merge-ready:
  >
  > 1. Forced no-media ablation: the retained forced solution is still {"rallies":[]}. That is an abstention, so the zero score does not yet demonstrate the intended no-media recall bound. Please rerun this
  >    diagnostic with an ablation-only instruction requiring at least one non-empty, schema-valid best-effort guess despite the missing media. This does not require changing the normal task instructions.
  >
  > 2. No-tools frame-dump ablation: supplying the complete selected 1 fps set (1,435 frames) is a reasonable sampling design; I am not asking for all 172,200 native frames. However, the retained trajectory
  >    used 44 shell/Python inspection calls, so it is not yet the requested no-tools condition. Please make that same selected frame set directly discoverable to the model—contact sheets or another pre-
  >    arranged presentation are fine—and rerun without task-inspection tool calls.
  >
  >    Please also correct the frame-dump summary: the run submitted/predicted 89 rallies, but the scorer matched 10 rallies and 7 strokes. It did not recover 89 of the 92 ground-truth rallies.
  >
  > A few smaller evidence/metadata items can be handled without starting a new full calibration campaign:
  >
  > - Antigravity parity: the retained run used the same six frozen task files; the only rules difference was the added sentence that a complete best-effort answer beats an empty one. Because the run
  >   nevertheless produced a non-empty answer and scored 0, I am comfortable accepting this disclosed one-line difference and am not requesting a strict-parity rerun. The rollout README says every retained
  >   harness includes a per-run net_guard log, but I only found the Codex and Claude logs. Please add the existing Antigravity log if it is recoverable; otherwise correct the claim and explicitly disclose that
  >   evidence gap.
  >
  > - Image identity: please record the actual image ID/digest used for the retained runs if it remains recoverable. If it is not recoverable, state that accurately and record the digest for the two new
  >   ablation runs and any future reruns. I am not asking for Codex or Claude to be rerun solely because the old digest was not recorded.
  >
  > - Provenance wording: as a small cleanup, describe the key as something like “machine truth from published structured annotations, with bounded video-audited terminal exceptions.” That is more precise than
  >   presenting the entire transformed key as purely machine-derived.
  >
  > Once the two ablations and the factual/evidence updates above are in place, this should only need a final merge recheck. I am not asking for reannotation, scorer redesign, native-rate frame extraction, or
  > another full three-harness campaign.

DDMeer:
@techgenmini  Thanks — all requested items are addressed at head ed414a9. Both requested ablations were rerun, and the six frozen task files remain byte-identical to the freeze commit.

One provenance note: the freeze commit was 1b7cbd5 when the runs were made. After rebasing onto upstream/main, the same frozen content is now reachable at 6aa6786. The old SHA remains in run-time metadata because that is what those runs actually recorded; scores.md documents the mapping.

Forced no-media now produces a non-abstaining best-effort answer: 96 rallies / 467 strokes, with 19/92 rallies and 54/387 strokes matched. Reward remains 0.000000.

For frame_dump_no_tools, all 1,435 selected 1 fps frames were pre-arranged into 30 contact sheets and passed directly as image input with all tools disabled. The retained trajectory has tool_use = 0; it submitted 43 rallies / 158 strokes and matched 3/92 rallies and 5/387 strokes, with reward 0.000000.

I also corrected the earlier frame-dump summary: 89 was the submitted-rally count, not recall. The old tools-based run matched 10 rallies and 7 strokes.

The Antigravity net_guard log is now committed, the image-identity wording has been narrowed to what the metadata actually supports, and the provenance wording has been updated as requested. The Antigravity parity caveat remains disclosed as previously accepted.

check_task.py passes 8/8, and the frozen six-file contract remains unchanged.

techgenmini:
> Hi @DDMeer — thanks. I completed the final re-review of live head ed414a9f2f1031a775f27a0d54a910dd912271e2. The prior bounded checklist is substantively closed.
  >
  > I verified the pre-/post-rebase freeze mapping (1b7cbd5 → 6aa6786) and confirmed that the six frozen task files remain byte-identical through the current head. I regenerated the 92-rally / 387-stroke key
  > and terminal audit with matching hashes, reran the verifier regressions and check_task.py (8/8), and independently regraded every retained final and ablation solution with exact agreement.
  >
  > The forced no-media run is now non-abstaining and still scores 0.000000. The replacement frame_dump_no_tools trajectory has zero tool_use calls, submits 43 rallies / 158 strokes, matches 3 rallies / 5
  > strokes, and scores 0.000000. The corrected earlier summary, Antigravity net-guard evidence, provenance wording, and narrowed image-identity claim close the corresponding prior items.
  >
  > One bounded evidence item remains before merge-ready:
  >
  > - The exact 30 contact sheets supplied to frame_dump_no_tools are not currently resolvable or reproducible from the retained artifacts. The repository contains their SHA-256 manifest, but not the image
  >   request payloads or a generation recipe; the runner reads pre-existing sheets from the contributor-local avb-abl-fdnotools:/workspace/sheets container. A hash manifest identifies the intended bytes, but
  >   without either the bytes or a reproducible derivation, a reviewer cannot verify the claimed 1,435-frame coverage, ordering, or presentation quality.
  >
  > The preferred low-bloat resolution is a deterministic sheet-generation script using the pinned source media and pinned tool/codec versions, with a final check that it reproduces every existing per-sheet
  > SHA-256. Alternatively, a revision-pinned archive is fine; please record its URL, revision, byte size, and whole-file SHA-256 while retaining the current per-sheet manifest. There is no need to commit the
  > JPEG binaries to this repository.
  >
  > If the existing sheets or their exact generation process are recoverable, no model rerun is needed. This is an evidence-chain correction, not a request to redesign the task or repeat calibration.
  >
  > Non-blocking factual cleanup: calibration/scores.md reports 150 Codex command_execution items, then says a looser count including file_change gives 63. I count 150 command executions plus five file-change
  > items, so that sentence should be corrected or removed.
  >
  > Once the frame-input evidence is resolvable, this should require only a focused artifact/head recheck. I am not requesting another calibration campaign, scorer change, or annotation update.

DDMeer:
@techgenmini Thanks — both items are addressed at head `9bc4340`. No model rerun was needed.

Contact-sheet reproducibility: `calibration/runpack/build_frame_dump_sheets.sh` is the deterministic derivation you asked for. It verifies the pinned media digest, extracts the full 1 fps / 1,435-frame set, tiles the frames into the same 30 7×7 contact sheets at 224×126 per tile, encodes JPEG at `-q:v 5`, and checks every regenerated sheet against the committed SHA-256 manifest.

A clean run from scratch reproduces all 30 committed digests exactly, so the frame coverage, ordering, and presentation are independently reproducible without committing the JPEG binaries.

The script also asserts ffmpeg `7.1.5-0+deb13u1` and fails on a mismatch. The Docker image ID is recorded for provenance but not pinned, since independent builds of the same Dockerfile do not necessarily produce the same image ID.

I also corrected the Codex turn-count note. The retained trajectory has 150 `command_execution` items, 5 `file_change` items, and 1 `todo_list` item. The previous “63” figure was stale and has been removed; the reported tool-call count remains 150 command executions.

`check_task.py` still passes 8/8, and the six frozen task files remain byte-identical through `9bc4340`.

Reviewer can reproduce the sheets with:

```bash
cd tasks/agentic_vbench_understanding/openttgames-rally-event-chain/calibration
./runpack/build_frame_dump_sheets.sh
```

Expected final line:

```text
all 30 sheets reproduce their committed digests
```

techgenmini:
> Hi @DDMeer — thanks. I completed a focused final recheck of live head 9bc4340ea39447b1ba9c50d6a868e3c2f21a3326. The previously remaining artifact-evidence item is closed.
  >
  > From a fresh Docker build, the pinned media digest passed, and build_frame_dump_sheets.sh enforced the expected ffmpeg version, extracted exactly 1,435 frames, generated 30 contact sheets, and reproduced
  > all 30 committed SHA-256 values. I also confirmed that the six frozen task files remain byte-identical to 6aa6786, ground-truth regeneration is byte-identical, all 13 verifier regressions pass, and
  > check_task.py passes 8/8. The corrected Codex counts also match the retained trajectory: 150 command_execution, 5 file_change, and 1 todo_list items.
  >
  > I see no remaining task, scorer, calibration, or artifact blocker, and no model rerun is needed.
  >
  > One non-blocking factual cleanup remains in the latest wall-time edit. scores.md says the values use the pre-/post-net_guard timestamps, but those timestamps produce approximately:
  >
  > - Codex: 35.5 min
  > - Antigravity: 40.6 min
  > - Claude: 101.0 min
  >
  > If the intended definition is instead run-metadata started through post-net_guard, the values are 34.5, 40.0, and 99.4 minutes. The current 34.5 / 40.6 / 99.3 row mixes those definitions. Please align the
  > sentence and table when convenient; this is documentation bookkeeping and does not require rerunning anything.
  >
  > With that non-blocking note, this is merge-ready from my review.

DDMeer:
@techgenmini  Thanks — fixed at `c5078b7`.

All three wall times now use the same pre-/post-`net_guard` definition:

| harness | wall time |
|---|---:|
| Codex | 35.5 min |
| Antigravity | 40.6 min |
| Claude | 101.0 min |

The Claude segment footnote has been aligned to the same definition as well. Rewards, turn counts, matched counts, and artifacts are unchanged. `check_task.py` still passes 8/8, and the six frozen task files remain byte-identical to `6aa6786`.

Thanks again for the careful review.

techgenmini:
Merge-ready

